PR #46, opened 2019-08-01
added new filter hook to single listing edit/back

themeaazz:


aazztech:
solve the conflict first please

PR #69, opened 2021-02-14
fix default search distance missing from shortcode searches

Terrox:
$_GET['miles'] is not always available from shortcodes with some settings combinations. I don't know about situations where you want a non-distance search, so those would need to be handled from the extra options in the search fields.


kowsar89:
From now on, we will accept pull request on alpha branch only

PR #84, opened 2021-03-17
added default lat, long for map data

sovware13:


kowsar89:
instead of long check, please use the 3rd parameter and send pull request again:

`get_directorist_option( 'default_latitude', '40.7127753', true );`

PR #86, opened 2021-03-18
fixed custom taxonomy issue with directory type

Armanul46:


kowsar89:
Why SearchForm.php file was updated? @Armanul46

PR #155, opened 2021-04-08
Fixing location map api field

syedgalib:


vairafiq:
Mybe it has some issue as I have tested it in user site.

kowsar89 (changes requested):
JS error: http://prntscr.com/11dhfht

kowsar89 (changes requested):
Style seems broken

kowsar89 (changes requested):
http://prntscr.com/11oyo7j

kowsar89 (approved):
Looks good to me

PR #163, opened 2021-04-12
fix tagline issue

vairafiq:


kowsar89 (changes requested):
http://prntscr.com/11ahwul

kowsar89 (changes requested):
issue still exists

PR #169, opened 2021-04-15
File field issue

Armanul46:


kowsar89 (changes requested):
1. http://prntscr.com/11ohb1s
2. Make same adjustment in legacy template

PR #173, opened 2021-04-18
Author profile page fixing

syedgalib:


kowsar89 (changes requested):
Pls dont add it in helper function

kowsar89 (changes requested):
Pls update code

PR #197, opened 2021-04-26
improve default icon and preview image in builder

vairafiq:


syedgalib (changes requested):
image


Store the argument in a local variable before modifying it.

syedgalib (dismissed):

kowsar89 (approved):

PR #198, opened 2021-04-27
remove confusing settings

vairafiq:


kowsar89 (changes requested):
http://prntscr.com/125i3rq

kowsar89 (approved):

PR #232, opened 2021-05-11
Remove legacy templates and codes

kowsar89:


vairafiq:
@kowsar89 better to undo changes from these files as galib is working on asset removing

image

PR #244, opened 2021-05-16
Fixing listings status issue with pricing plan while no plan is assigned

syedgalib:


kowsar89 (commented):
Why 'metadata_exists' func used here?

syedgalib:
To check if the meta key is set or not. The meta key "_featured" and "listings_status" are mandatory and they are set conditionally based on the various situations.

But during backend listing submission, if the pricing plan extension is enabled and no plan is assigned to the listing, the meta keys get skipped. As a result, the published listing doesn't appear on the front end.

So I have added a check at the bottom of the listing saving process so that the meta keys get stored to the listing with a default value in case they get skipped.

vairafiq (approved):
This code should be moved to DPP & WPP. 
It's OK for now

PR #245, opened 2021-05-18
Single selection now works in Form Tag

kowsar89:


vairafiq (changes requested):
@kowsar89 https://prnt.sc/132awe5

PR #253, opened 2021-05-18
added filter hook to range slider

syedgalib:


kowsar89 (approved):
works fine from my side

PR #319, opened 2021-06-30
fix contact listing form not working issue

vairafiq:


kowsar89 (approved):
Works fine

includes/model/SingleListing.php
```diff
@@ -546,6 +546,13 @@ public function price_range_html() {
 		return $result;
 	}
 
+	public function hide_contact_owner_form() {
```
obiPlabon: How about if we write the function name in a more conditional fashion and with a little bit of documentation...

```php
/**
 * Check if the contact owner form is set to hide or not.
 *
 * @return bool
 */
public function is_contact_owner_form_hidden() {
    return (bool) get_post_meta( $this->id, '_hide_contact_owner', true );
}
```

obiPlabon (approved):

PR #321, opened 2021-06-30
fix default preview image issue when multi directory is disabled

vairafiq:


kowsar89:
Closed due to issue generated from somewhere else

PR #330, opened 2021-07-01
HTML end tag Fix in single templates

HeyMehedi:


kowsar89:
Testing

HeyMehedi:
Update: some hardcoded to translatable strings

kowsar89:
closed in the favor of #386

PR #335, opened 2021-07-05
Fix/listing url issue at email template

syedgalib:


vairafiq (changes requested):
Need to adjust the changes in email helper as well

vairafiq (approved):
Looks OK

obiPlabon (approved):

PR #336, opened 2021-07-05
Fix/rank math seo issue

syedgalib:


includes/helper-functions.php
```diff
@@ -7536,6 +7536,16 @@ function atbdp_disable_overwrite_yoast() {
     atbdp_can_overwrite_yoast();
 }
 
+if ( ! function_exists( 'directorist_is_active_rankmath' ) ) {
```
obiPlabon, line 7539: `class_exists( 'RankMath' )` by default returns boolean, so we don't need to check and return `true/false` explicitly.

```php
function directorist_is_active_rankmath() {
    return class_exists( 'RankMath' );
}

```

includes/classes/class-seo.php
```diff
@@ -36,16 +36,34 @@ public function __construct()
             // --------------------------------------------
             // Meta Title
             add_filter( 'rank_math/frontend/title', function( $title ) {
```
obiPlabon, line 38: Anonymous functions are not unhookable or cannot be removed using `remove_filter` and `remove_action`. Could you please make them a regular function/method?

obiPlabon (changes requested):

vairafiq (approved):

obiPlabon (approved):

obiPlabon:
Thank you @syedgalib bhai

PR #337, opened 2021-07-06
Changes the  hook name of login form settings

Armanul46:


obiPlabon (commented):
Since you are renaming the hook name then why the prefix is still `atbdp_`? Aren't we supposed to use `directorist_`?

Note:
I'd suggest not to change or rename any action hooks, filter hooks, constants, api functions, and classes names without fallback and deprication notice. Very soon we are going to expose public APIs for 3rd party integration. So we must develop this habbit.

vairafiq (approved):

obiPlabon (approved):

obiPlabon:
Thank you @Armanul46 bhai

PR #349, opened 2021-07-08
Fix/category fields

vairafiq:


kowsar89 (changes requested):
If custom field is on different section, it doesn't work as expected

obiPlabon (approved):

kowsar89 (approved):

PR #359, opened 2021-07-13
Remove/pure select

syedgalib:


vairafiq (approved):
Tested and works fine

obiPlabon (approved):
Pass

PR #382, opened 2021-07-28
Fix/reset filter is not working in nested situation

syedgalib:


kowsar89:
@syedgalib pls fix conflict

PR #397, opened 2021-08-03
fixed select2 dropdown menu issue

syedgalib:


vairafiq (approved):
Tested and fixed the issue. We will remove select2 and apply custom select2 in an upcoming release

PR #423, opened 2021-08-23
Directorist team issue

Armanul46:
fixed sorting issue of all authors


vairafiq (approved):
Tested and works fine

vairafiq (changes requested):
Template should not have any logic

PR #436, opened 2021-08-30
Update/Appsero client lib

obiPlabon:
Updated appsero client library as the updated library included the plugin tracking feature by default. And removed the old deactivation reason popup.


vairafiq (approved):
reviewed

kowsar89 (approved):
pass

PR #455, opened 2021-09-08
reset query

vairafiq:


obiPlabon (commented):
Why do we have to reset the query here?

HeyMehedi (dismissed):

obiPlabon (approved):
Okay

directorist-base.php
```diff
@@ -786,6 +786,7 @@ public function get_popular_listings($count = 5, $listing_id = '')
 						}
```
kowsar89, line 787: `wp_reset_postdata()` could go just after while loop

directorist-base.php
```diff
@@ -825,6 +826,7 @@ public function get_popular_listings($count = 5, $listing_id = '')
...
```
kowsar89, line 827: `wp_reset_postdata() `could go just after while loop

directorist-base.php
```diff
@@ -844,6 +846,7 @@ public function get_popular_listings($count = 5, $listing_id = '')
...
 
 					}
```
kowsar89, line 848: `wp_reset_postdata()` could go just after while loop

directorist-base.php
```diff
@@ -879,6 +882,7 @@ public function get_popular_listings($count = 5, $listing_id = '')
...
```
kowsar89, line 884: `wp_reset_postdata() `could go just after while loop

includes/model/ListingDashboard.php
```diff
@@ -421,7 +421,7 @@ public function get_announcements() {
 				'content' => $announcement->post_content,
 			];
 		}
-
+
```
kowsar89, line 424: unnecessary space, this file shouldn't have any changes

kowsar89 (changes requested):

kowsar89 (approved):
Looks good

obiPlabon (approved):
Okay

PR #460, opened 2021-09-13
fix select field issue

vairafiq:
ID attribute causes html validation issue and also has some issue with select2


obiPlabon (approved):
Reviewed

kowsar89 (approved):
seems ok

PR #466, opened 2021-09-16
fix uploader issue

vairafiq:
It has a conflict with pricing plan extension, assigning image limit from build and plan will cause the issue


kowsar89 (approved):
Tested, looks good

PR #470, opened 2021-09-20
fix plural typo

vairafiq:


obiPlabon:
image
There's only one search form then why should we call it search forms?

obiPlabon (approved):
Reviewed

PR #480, opened 2021-09-29
fixed search listing shortcode atts not working after ajax load

syedgalib:


vairafiq (approved):
Tested. Works fine

includes/classes/class-shortcode.php
```diff
@@ -179,6 +179,9 @@ public function all_locations( $atts ) {
 	}
 
 	public function search_listing( $atts ) {
+
+		// var_dump(  $atts );
```
kowsar89, line 183: remove var_dump

kowsar89 (changes requested):

kowsar89 (changes requested):
Doesn't work if shortcode is used multiple times in a page.
image

syedgalib:
fixed

PR #489, opened 2021-10-10
apply filter - directorist_delete_listings_from_dashboard

MahfuzulAlam:
Introduce a hook "directorist_delete_listings_from_dashboard" to determine the status of the deleted listings from the dashboard whether it is going to trash or deleted forever.

https://i.imgur.com/oU5M5L7.png

@return 'delete' to delete forever or 'anything' to send the listing to the trash


MahfuzulAlam (commented):
This filter will help the user to overwrite the default social link options.

kowsar89 (changes requested):
1. Please rename the following hooks:
```
directorist_default_social_links => directorist_social_items
directorist_delete_listings_from_dashboard => directorist_listing_delete_or_trash
```
2. Fix merge conflict
@MahfuzulAlam

vairafiq (dismissed):
Looks good

includes/model/ListingDashboard.php
```diff
@@ -68,12 +68,13 @@ public function ajax_listing_tab() {
 	}
 
 	public function listing_task( $task, $taskdata ){
```
kowsar89, line 70: $task variable isn't used anywhere in the code!

kowsar89 (changes requested):

vairafiq (approved):
tested

includes/model/ListingDashboard.php
```diff
@@ -68,12 +68,13 @@ public function ajax_listing_tab() {
...
-		if ( $task == 'delete' ) {
-
-			do_action( 'directorist_listing_deleted', $taskdata );
-
+		$task = apply_filters('directorist_listing_delete_or_trash', $task);
+		if( $task == 'delete' ){
 			wp_delete_post( $taskdata );
+		}else{
```
kowsar89, line 74: $task can be anything in future- like 'delete', 'update' etc, so "else" command should not be used here to avoid bugs in future. Only 'if' is allowed

kowsar89 (changes requested):

PR #491, opened 2021-10-10
fixed multi directory nav redirects to blog from homepage

syedgalib:


vairafiq (approved):
Tested and works fine

PR #502, opened 2021-10-18
Fix/builder widget label issue

syedgalib:


kowsar89 (approved):
Works fine from my side

PR #503, opened 2021-10-19
Fix/map data persing

vairafiq:


vairafiq:
some console error 
image

kowsar89 (approved):
seems ok

PR #507, opened 2021-10-20
Fix dashboard dashboard bookmark issue

vairafiq:
There was an issue with unfavourite a listing from user dashboard page

image


kowsar89 (approved):
works fine

PR #521, opened 2021-10-26
New/updated review

vairafiq:


kowsar89 (changes requested):
image

kowsar89 (changes requested):
Review/Reply button doesn't work in guest/incognito mode

kowsar89:
image

kowsar89:
image

kowsar89 (changes requested):
There should be option for enabling/disabling reply feature, and default should be disabled for compatibility

kowsar89 (approved):
Works fine

PR #528, opened 2021-10-31
fix issue with polylang rewrite roles

vairafiq:
User issue: Polylang categories and locations do not work properly, an issue for 2 years now, apart from that the Directorist is polylang compatible.

Issue found when someone change language https://prnt.sc/1xwlbjm

URL : https://superclassics.eu/language/nl/single-catogery/?category=abarth


kowsar89 (approved):
seems ok

PR #534, opened 2021-11-01
Select custom field issue on seach result page

Armanul46:


Armanul46:
https://app.clickup.com/t/dzjuvt

templates/search-form/custom-fields/select.php
```diff
@@ -22,7 +22,7 @@
 
 			<?php
 			foreach ( $data['options'] as $option ) {
-				printf('<option value="%s"%s>%s</option>', $option['option_value'], selected(  $value === $option[ 'option_value' ] ), $option['option_label']);
+				printf('<option value="%s"%s>%s</option>', $option['option_value'], selected(  $value === $option[ 'option_value' ], true, false ), $option['option_label']);
```
obiPlabon, line 25: Missing proper sanitization `esc_attr`, `esc_html` in option label and value. Also, maybe you can remove unnecessary formatted code `printf`.

Also, you didn't use `selected()` in an expected way.

I've rewritten the whole template for your inspiration.

```php
$field_id = uniqid( 'directorist-input-' );

?>
<div class="directorist-search-field">
	<?php if ( ! empty( $data['label'] ) ) : ?>
		<label for="<?php echo $field_id; ?>"><?php echo esc_html( $data['label'] ); ?></label>
	<?php endif; ?>

	<div class="directorist-select">
		<select id="<?php echo $field_id; ?>" name="custom_field[<?php echo esc_attr( $data['field_key'] ); ?>]" <?php echo ! empty( $data['required'] ) ? 'required="required"' : ''; ?> data-isSearch="true">
			<option value=""><?php esc_html_e( 'Select', 'directorist' )?></option>

			<?php foreach ( $data['options'] as $option ) : ?>
				<option value="<?php echo esc_attr( $option['option_value'] ); ?>" <?php selected( $value, $option[ 'option_value' ] ); ?>><?php echo esc_html( $option['option_label'] ); ?></option>
			<?php endforeach; ?>
		</select>
	</div>
</div>
```

obiPlabon (changes requested):

PR #541, opened 2021-11-04
Updated rest api listings endpoint

obiPlabon:
- Rating query
- Price query
- Price range query
- Expired query
- Listings status query fix
- Radius query


vairafiq (approved):
pass

kowsar89 (approved):

PR #553, opened 2021-11-10
add data to pricing template

vairafiq:


kowsar89:
Will use alternative way instead of using these data

PR #571, opened 2021-11-16
Fix security issues

obiPlabon:


kowsar89 (changes requested):
template version not updated

includes/classes/class-user.php
```diff
@@ -1,264 +1,293 @@
 <?php
 /**
- * User class of Directorist
+ * Directorist user class.
  *
- * This class is for interacting database table
+ * User registration and acess management related functionalities can be found in this class.
```
kowsar89, line 5: Typo

kowsar89 (changes requested):

includes/classes/class-user.php
obiPlabon: Fixed

kowsar89 (approved):
Works fine

vairafiq (approved):

PR #593, opened 2021-11-28
Listings table refactoring

obiPlabon:
Screenshot 2021-11-28 at 5 06 26 PM

## Changes
- Renamed "Listing Name" to "Name" since it's obvious that it's a listings table.
- Renamed "Directory Type" to "Directory" - actually that's what a directory type is!
- Updated the "Featured" icon and removed the non-featured icon. After all, why do we need an icon for non-exclusive items!
- Capitalized listing statuses
- Author link is now used for author filtering instead of the front-end author archive page.
- Removed category icon and directory icon. It's not that much important for the listings table. And reduced some queries!
- Used terms caching for locations and categories. 50% query reduced due to caching!!!
- And refactored some codes...


vairafiq (approved):
Reviewed

PR #601, opened 2021-12-05
Improve/select option fields

syedgalib:
- Added - Lazy loading feature to category and location field.
- Added - Option in settings panel to toggle lazy loading the category and location field.
- Added - A new action hook called "directorist_installed" that runs one time after the plugin gets installed.
- Added - A new action hook called "directorist_updated" that runs one time after the plugin gets updated.
- Added - A new helper function called "directorist_has_no_listing" that tells is site has any listings or not


obiPlabon:
`directorist_on_mount` - conventionally **mount** is not suitable for WordPress, instead we can simply use `directorist_installed`. That's easy to remember even for a novice WordPress dev, `on_mount` or `mount` is good for someone with React or Vue or JS experience.

includes/classes/class-script-helper.php
```diff
@@ -132,12 +132,13 @@ public static function get_listings_data() {
...
         $review_approval = get_directorist_option( 'review_approval_text', __( 'Your review has been received. It requires admin approval to publish.', 'directorist' ) );
         $enable_reviewer_content = get_directorist_option( 'enable_reviewer_content', 1 );
-        
+
         $data = array(
             'nonce'                       => wp_create_nonce( 'atbdp_nonce_action_js' ),
             'directorist_nonce'           => wp_create_nonce( directorist_get_nonce_key() ),
             'ajax_nonce'                  => wp_create_nonce( 'bdas_ajax_nonce' ),
             'ajaxurl'                     => admin_url( 'admin-ajax.php' ),
+            'home_url'                    => home_url(),
```
obiPlabon, line 141: Instead of using the `home_url()` we should use `rest_url()`, as the rest URL can be modified using filter hook.

More details...
https://developer.wordpress.org/reference/functions/rest_url/
https://developer.wordpress.org/reference/functions/rest_get_url_prefix/

includes/helper-functions.php
```diff
@@ -8574,3 +8580,15 @@ function directorist_get_supported_file_types() {
 		return array_merge( $carry, $group );
 	}, [] );
 }
+
+
+function directorist_is_new_user() {
```
obiPlabon, line 8585: The function name is kinda misleading, maybe we can use `directorist_is_listings_empty()`? Also, we can optimize queries by disabling caching and pagination to avoid worst case.
```
'no_found_rows' => true 
'cache_results' => false
`

obiPlabon (changes requested):

kowsar89 (approved):
Pass. Seach-form seems working though

obiPlabon:
image
These's an encoding issue...

obiPlabon (approved):
Looks good

vairafiq (approved):
pass

kowsar89 (approved):
pass

PR #604, opened 2021-12-06
Fixed display_preview_image parameter issue of all listings shortcode

Armanul46:
https://prnt.sc/21yimmb


kowsar89 (changes requested):
Doesn't work when disabled from settings: http://prntscr.com/21yz3u8

kowsar89 (approved):
works fine

PR #608, opened 2021-12-08
Fixed parameter issue of search listing shortcode

Armanul46:
https://prnt.sc/22a2hqz


vairafiq (changes requested):
image
sometimes it might be empty!

includes/helpers/trait-uri-helper.php
```diff
@@ -89,4 +89,13 @@ public static function get_theme_template_path_for( $template ) {
 			return $index_path;
 		}
 	}
+
+	public static function is_field_allowed_in_atts( $widget_name, $atts = [] ) {
```
kowsar89, line 93: This file is intended for URL related functions only.
pls move this func into 'Directorist_Listing_Search_Form' class

kowsar89 (changes requested):

kowsar89:
Discussion summery:
1. Attribute should be like 'filter_category' instead of 'category'
2. Should also work for 'all listings' and 'search result' page
3. After task is completed, update doc

includes/model/Listings.php
```diff
@@ -1931,7 +1931,7 @@ public function viewas_dropdown_template() {
 		public function search_form_template() {
 			$args = array(
 				'listings'   => $this,
-				'searchform' => new Directorist_Listing_Search_Form( $this->type, $this->current_listing_type ),
+				'searchform' => new Directorist_Listing_Search_Form( $this->type, $this->current_listing_type, $this->atts ),
```
kowsar89, line 1934: All Liistings atts should not be polluted with Search_Form atts, both are different

kowsar89 (changes requested):
Empty space

includes/model/SearchForm.php
```diff
@@ -341,9 +341,22 @@ public function get_pricing_type() {
...
+		if ( $atts && ( 'no' == $atts ) ){
+			return false;
+		}
+		return true;
+	}
+
 	public function field_template( $field_data ) {
 		$key = $field_data['field_key'];
 
+		if( ! $this->is_field_allowed_in_atts( $field_data['widget_name'], $this->atts ) ) {
```
kowsar89, line 356: why pass $this->atts parameter? since the function is in the same class that value could be retrieved easily without passing parameter

includes/model/Listings.php
```diff
@@ -1929,9 +1929,10 @@ public function viewas_dropdown_template() {
 		}
 
 		public function search_form_template() {
+			$search_field_atts = array_intersect_key( $this->atts, array_flip( preg_grep("/^filter_/", array_keys( $this->atts ) ) ) );
```
kowsar89, line 1932: Please make it more readable. eg you can make it into multiple lines, optionally add comments
Ref: https://www.freecodecamp.org/news/long-code-vs-short-code/

kowsar89 (changes requested):

kowsar89:
Related: https://github.com/sovware/directorist/pull/616

PR #615, opened 2021-12-23
Fix/theme update

vairafiq:
Issues: https://app.clickup.com/t/1x0gzu7


kowsar89 (changes requested):
not working

kowsar89 (changes requested):
Doesn't work for onelisting(free) and drealestate theme

vairafiq:
Fixed, Remote API updated

kowsar89 (approved):
Works now

PR #624, opened 2021-12-29
fix wrong type assign if search form used with listing form

vairafiq:
Issue: https://app.clickup.com/t/1xby3dy


obiPlabon (approved):
Working as expected.

PR #627, opened 2022-01-03
Removed style from body tag

devShamim:
Removed style from body tag and some unnecessary scss files.
Added flex-wrap to the directory type list.

Issue:
https://wordpress.org/support/topic/dont-use-if-you-care-about-usability-and-are-not-using-their-theme/


vairafiq:
image
why the styles are removed?

shamimsaj:
> image why the styles are removed?

Those files were empty and useless.

vairafiq (dismissed):

vairafiq (approved):
Tested

PR #634, opened 2022-01-11
added js hooking feature

syedgalib:
Hook Uses

### Filter Hook
```js
directoristHooks.addFilter( 'hookName',  function( args ) {
  // Do Something
  return args;
});

directoristHooks.applyFilter( 'hookName', args );
```

### Action Hook
```js
directoristHooks.addAction( 'hookName',  function( args ) {
  // Do Something
});

directoristHooks.doAction( 'hookName', args );
```


obiPlabon:
Why do we need this? Besides WP already have a package for this https://www.npmjs.com/package/@wordpress/hooks we can use that.

PR #639, opened 2022-01-13
Fix old review system bc

obiPlabon:
Compatible class and methods.

```
ATBDP_Review_Rating
- print_static_rating()
- get_average()

ATBDP_Review_Rating_DB
- count()
- get_ratings_by()
```


vairafiq (approved):
Works fine

Issue #444, opened 2021-09-05
External links from settings should open in new tab


kowsar89:
Screenshot:

![image](https://user-images.githubusercontent.com/6196149/132113945-c095cfb1-cabc-4021-a1ae-5ed12ebff898.png)

PR #640, opened 2022-01-16
External links from settings & builder open in new tab

Armanul46:
https://github.com/sovware/directorist/issues/444


vairafiq (approved):
Working

PR #645, opened 2022-01-18
Rename hook

vairafiq:
Issue: Extra space before review tab in builder https://prnt.sc/26ezey7

-- The hook is used in builder and settings panel layout, theme team has confirmed that this hook is only used for extending settings layout, not for builder. So it is safe to rename the builder hook.


kowsar89 (approved):
looks good

PR #654, opened 2022-01-24
Add: Category Color Meta

HeyMehedi:


vairafiq:
Meta has no uses. Please use this color in all category page

PR #663, opened 2022-01-27
Fix map popup showing default 'grid' image if no image found rather than using image set on directory builder

M4hd1BD:
Fix map popup showing default 'grid' image if no image found rather than using an image set on directory builder


RabbiIslamRony:
If the issue still exists, we can reopen it. For now, I’m closing it since it’s an old PR

PR #665, opened 2022-01-27
Fix an issue with user not able to select address if there's no map added on the add listing form

M4hd1BD:
Fix an issue with the user not being able to select address if there's no map added on the add listing form
https://www.loom.com/share/47a975740e40404e8c6a562d9cd5bc4c


vairafiq (approved):
Tested works fine

PR #669, opened 2022-02-03
add support for new user role 'seller'

vairafiq:
issue: https://app.clickup.com/t/1zauxu3


obiPlabon (approved):
Tested on users' site.

PR #673, opened 2022-02-07
Fix an issue with the user not being able to select address if there's no map added on the add listing form

M4hd1BD:
Fix an issue with the user not being able to select address if there's no map added on the add listing form


vairafiq:
not working, make sure you have nun `npm run dev`

M4hd1BD:
@vairafiq fixed issue, kindy recheck

vairafiq (approved):
Working now

Issue #672, opened 2022-02-07
PHP warning on fresh installation


obiPlabon:
Found this warning in a full fresh installation without any data imported.

```php
Warning: Invalid argument supplied for foreach() in /app/wp-content/plugins/directorist/includes/classes/class-tools.php on line 281
```

PR #674, opened 2022-02-07
Refactored ATBDP_Tools class

obiPlabon:
- Allow execution on the admin side only using `is_admin`.
- PHP warning invalid foreach argument - #672 
- Import capability check for ajax listing import.


vairafiq (approved):
Tested, works fine

PR #679, opened 2022-02-07
Add WP Date Format

HeyMehedi:


vairafiq:
Template version is not updated

obiPlabon:
Use https://developer.wordpress.org/reference/functions/date_i18n/

PR #688, opened 2022-02-10
fix console error on listing submission

vairafiq:
**Fixed**
- There are few console error
- Directory type has not been assigned


obiPlabon (approved):
Working

PR #689, opened 2022-02-10
added the missing js code of category-custom-field issue

Armanul46:


vairafiq:
not working if Lazy load category and location fields is enabled from settings

PR #725, opened 2022-02-22
Ajax search

Armanul46:
please enable "Instance Search" before testing
https://prnt.sc/ctqs5UTQfOef


obiPlabon (changes requested):
Please update all the references of "instance search, Instance search, ajax search, Ajax search" to "instant search, Instant search". Definitely including CSS class, js function name, PHP class, PHP function, nonce etc. The term must be consistent.

obiPlabon:
image
Multiple tags are selected but only the last tag is used in the url.

kowsar89 (approved):
Checked only models and template code changes, looks ok

obiPlabon (approved):
Search functionality looks good and working as expected.

includes/model/Listings.php
```diff
@@ -1616,6 +1620,18 @@ public function loop_wrapper_class() {
 			return implode( ' ' , $class );
 		}
 
+		public function instant_searching_class() {
```
obiPlabon, line 1623:
```php
/**
 * Displays the class names for the listings wrapper element.
 *
 * @since 7.2.0
 *
 * @param string|string[] $class Space-separated string or array of class names to add to the class list.
 */
public function wrapper_class( $class = '' ) {
	// Separates class names with a single space, collates class names for wrapper tag element.
	echo 'class="' . esc_attr( implode( ' ', $this->get_wrapper_class( $class ) ) ) . '"';
}

/**
 * Retrieves an array of the class names for the listings wrapper element.
 *
 * @since 7.2.0
 *
 * @param string|string[] $class Space-separated string or array of class names to add to the class list.
 * @return string[] Array of class names.
 */
public function get_wrapper_class( $class = '' ) {
	$classes = array(
		'directorist-archive-contents',
	);

	if ( 'yes' === $this->instant_search ) {
		$classes[] = 'directorist-instant-search';
	}

	if ( ! empty( $class ) ) {
		if ( ! is_array( $class ) ) {
			$class = preg_split( '#\s+#', $class );
		}
		$classes = array_merge( $classes, $class );
	} else {
		// Ensure that we always coerce class to being an array.
		$class = array();
	}

	$classes = array_map( 'esc_attr', $classes );

	/**
	 * Filters the list of CSS listings wrapper class names for the wrapper.
	 *
	 * @since 7.2.0
	 *
	 * @param string[] $classes An array of listings wrapper class names.
	 * @param string[] $class   An array of additional class names added to the listings wrapper.
	 * @param object   $this    An instantce of Directorist_Listings
	 */
	$classes = apply_filters( 'directorist_listings_wrapper_class', $classes, $class, $this );

	return array_unique( $classes );
}
```

obiPlabon (changes requested):

includes/model/Listings.php
obiPlabon: Inspired by https://developer.wordpress.org/reference/functions/body_class/

PR #733, opened 2022-02-27
Refactor user favorites

obiPlabon:
added utility to handle user favorites. needed to do this for GamiPress, BuddyPress, and BuddyBoss.


kowsar89 (approved):
pass

vairafiq (approved):
Tested

Issue #757, opened 2022-03-12
Email does not reach to Admin before approving the review. After approving the review admin receives the email


vairafiq:
https://lobby.sovware.com/wp-admin/admin.php?page=pm_projects#/projects/8/kanboard/tasks/1562

PR #761, opened 2022-03-13
fix: review email issue #757

obiPlabon:


vairafiq (approved):
Checked in new environment and worked perfectly

PR #764, opened 2022-03-14
fi an issue with Preview Image not setting if importing one image (it gets to slider instead)

M4hd1BD:
fi an issue with Preview Image not setting if importing one image (it gets to slider instead)


vairafiq (approved):
Tested looks good

PR #765, opened 2022-03-14
introduce a hook to do stuff with each listing after importing them

M4hd1BD:


includes/classes/class-tools.php
```diff
@@ -204,7 +204,7 @@ public function atbdp_import_listing()
 
                         update_post_meta($post_id, '_listing_img', $attachment_ids );
                     }
-
+                    do_action('directorist_after_import_single_listing', $post_id, $post);
```
obiPlabon, line 207: Could you please rename it to `directorist_listing_imported`?

```php
/**
 * Fire this event once a listing is successfully imported from CSV.
 *
 * @param int $post_id Listing id.
 * @param array $post  Listing data.
 */
do_action( 'directorist_listing_imported', $post_id, $post );
```

obiPlabon (changes requested):

includes/classes/class-tools.php
M4hd1BD: adjusted

vairafiq:
Screen Shot 2022-03-16 at 3 06 08 PM

@since is missing in docblock

vairafiq (approved):
Looks good

PR #766, opened 2022-03-14
Optimize/asset

syedgalib:


views/admin-templates/post-types-manager/edit-listing-type.php
```diff
@@ -2,14 +2,16 @@
     <?php 
         $enable_multi_directory = get_directorist_option( 'enable_multi_directory', false );
         $enable_multi_directory = atbdp_is_truthy( $enable_multi_directory );
+
+        $builder_data = base64_encode( json_encode( $data['directory_builder_data'] ) );
```
kowsar89, line 6: pls avoid `base64_encode`, using only `json_encode` should do the job

kowsar89 (changes requested):

PR #769, opened 2022-03-16
Refactor listings model and template for listings wrapper class

obiPlabon:
Inspired by https://developer.wordpress.org/reference/functions/body_class/


kowsar89 (approved):
seems ok

PR #792, opened 2022-03-24
Zipcode radius search

Armanul46:


vairafiq:
image

Need to solve conflict

assets/src/js/public/search-form-listing.js
```diff
@@ -192,6 +192,65 @@
         }
     });
 
+    $('body').on("keyup", '.zip-radius-search', function(){
```
obiPlabon, line 195: Use debounce mechanism here, otherwise, there'll be lots of unnecessary API calls.
https://davidwalsh.name/javascript-debounce-function

assets/src/js/public/search-form-listing.js
```diff
@@ -192,6 +192,65 @@
...
+        let res = '';
+        $.ajax({
+            url: `https://nominatim.openstreetmap.org/?postalcode=+${zipcode}+&format=json&addressdetails=1`,
+            type: "POST",
+            data: {},
+            success: function( data ) {
+                if( data.length === 1 ) {
+                    var lat = data[0].lat;
+                    var lon = data[0].lon;
+                    $('#zip-cityLat').val(lat);
```
obiPlabon, line 216: Please use a class selector instead of ID. make sure multiple search form works perfectly on a single page.

assets/src/js/public/search-form-listing.js
```diff
@@ -383,5 +442,11 @@
                 });
         }
     }
+    document.querySelectorAll(".directorist-search-contents").forEach(function (el) {
```
obiPlabon, line 445: Use jQuery instead

includes/modules/multi-directory-setup/class-multi-directory-manager.php
```diff
@@ -2970,6 +2970,15 @@ public function prepare_settings() {
                                     [ 'value' => 'kilometers', 'label' => 'Kilometers' ],
                                 ]
                             ],
+                            'radius_search_based_on' => [
```
obiPlabon, line 2973: Do we have a radio button? Use that if available.

obiPlabon (changes requested):
@Armanul46 please use CSS classes instead of ids. Use jQuery instead of DOM APIs.

obiPlabon (approved):
working fine

PR #804, opened 2022-03-27
Fix blocks issues

obiPlabon:
- Block wasn't listed on wp.org plugin read, so added `block.json` schema file.
- Fix block sorting issue


vairafiq (approved):
pass

PR #810, opened 2022-03-28
fix location & category searching issue in search result ( instant search )

Armanul46:


vairafiq:
Screen Shot 2022-03-28 at 3 29 40 PM
After selecting a category its searching but no result found

vairafiq (approved):
Checked

PR #814, opened 2022-03-29
fixed select2 does not work after re-initialization

syedgalib:


vairafiq:
Category field back to searched category when select2 lazy loading in enabled https://www.loom.com/share/1321196be15f454a946c979587f728a3

vairafiq (approved):
Working

Issue #827, opened 2022-03-31
single taxonomy shows listings from all taxonomy if directory type parameter present in the url


M4hd1BD:
https://prnt.sc/tktsoUddYoPM

PR #831, opened 2022-04-03
single taxonomy shows listings from all taxonomy if directory type parameter present in the url

Armanul46:
https://github.com/sovware/directorist/issues/827


vairafiq (approved):
tested

Issue #820, opened 2022-03-30
Popular Categories Not Working


vairafiq:
Private Link: https://app.clickup.com/t/20k2ann

RabbiIslamRony:
Confirmed—the underlying issue is still present. Categories are limited before they are filtered by directory type, which can produce missing or incomplete Popular Categories results. PR #834 contained the intended approach but was closed without being merged. This issue should remain open.

RabbiIslamRony:
A focused fix is ready in #2950. Directory filtering now happens inside the term query before the configured limit, with support for both indexed and legacy directory metadata. LocalWP verification covered competing high-count categories across directories, root/child filtering, empty terms, duplicate prevention, and the bounded SQL query.

PR #834, opened 2022-04-04
Popular category issue in search form

Armanul46:
https://github.com/sovware/directorist/issues/820


vairafiq:
It might cause performance issue. `number` should be restricted in main query and better to go with meta query

Issue #800, opened 2022-03-24
Membership - Anyone can register - Not Respected


dimalif:
**Anyone can register** Wordpress option is not respected. 

While we selected not to allow registrations for the moment, it seems anyone can register.

![Screenshot](https://user-images.githubusercontent.com/53921528/159942972-e19c7cf2-63a2-4e2d-ade9-7e77d7a7f0d5.png)

vairafiq:
@dimalif could you please check the discussion on #840

PR #840, opened 2022-04-07
improve new user registration

vairafiq:
Issue https://github.com/sovware/directorist/issues/800

### DIFF
--- a/includes/model/Account.php
+++ b/includes/model/Account.php
@@ -122,7 +122,7 @@ class Directorist_Account {
 				'author_checked'	   => ( 'general' != $user_type ) ? 'checked' : '',
 				'general_checked'	   => ( 'general' == $user_type ) ? 'checked' : '',
 			);
-
+			ob_start();
 			if ( ! empty( $atts['shortcode'] ) ) { Helper::add_shortcode_comment( $atts['shortcode'] ); }
 			echo Helper::get_template_contents( 'account/registration', $args );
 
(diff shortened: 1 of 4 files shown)


obiPlabon (commented):
We cannot simply add this check without informing the users beforehand. As the `user_can_register` is disabled by default on all the WordPress installations. So if we push this without notifying the users then there's a chance that users will not accept this change easily.

vairafiq:
Screen Shot 2022-04-10 at 10 16 19 AM

Great! Then should we push this warning to Directorist >> System Status?

vairafiq:
Screen Shot 2022-04-11 at 10 15 10 AM

New option introduced

includes/model/Account.php
```diff
@@ -35,14 +35,44 @@ public function render_shortcode_login( $atts = [] ) {
...
 		ob_start();
 		if ( ! empty( $atts['shortcode'] ) ) { Helper::add_shortcode_comment( $atts['shortcode'] ); }
-		echo Helper::get_template_contents( 'account/login' );
+		echo Helper::get_template_contents( 'account/login', $args );
 
 		return ob_get_clean();
 	}
 
 	public function render_shortcode_registration( $atts ) {
+
```
obiPlabon, line 75: if ( ! get_directorist_option( 'new_user_registration', true ) ) {
    return;
}

obiPlabon (approved):

Issue #786, opened 2022-03-22
Directory Name Creation issue, containing special characters


dimalif:
Trying to create a new directory with the builder, named "Cars & Vehicles", i get "directory name exists" in red. 

Despite the red error, the directory is created. 

Apparently the "&" is messing things up. I haven't tested other special symbols.

syedgalib:
Fixed @ https://github.com/sovware/directorist/pull/845

PR #845, opened 2022-04-11
Fix/directory name creation issue with special character

syedgalib:
https://github.com/sovware/directorist/issues/786


vairafiq (approved):
Working as expected

Issue #781, opened 2022-03-20
Select2 category field is not working when search widget used in sidebar


vairafiq:
https://prnt.sc/UvTN_Kp67n8j

syedgalib:
Fixed #846

PR #846, opened 2022-04-12
select2 category field is not working when search widget used in sidebar

syedgalib:
https://github.com/sovware/directorist/issues/781


vairafiq (approved):
Looks good now

PR #888, opened 2022-05-17
Selected listings query with instant search

obiPlabon:
When the instant search is enabled and some specific listings ids are passed to all listings shortcode then the ajax query doesn't work on the selected listings instead it loads all the listings.


vairafiq (approved):
Checked

PR #890, opened 2022-05-18
changed the child category expander to chevron

M4hd1BD:


tanjimhasan (commented):
Need to change variable name also and icon should be replaced by icon not letters. @devShamim will update those icons at assets-optimization  branch. @devShamim Please take a note.

Issue #915, opened 2022-05-31
Listings archive card view issue with custom fields


obiPlabon:
1. When checkbox custom field added but the options are empty.
2. Empty list items when custom field added but there no saved value.
<img width="528" alt="image" src="https://user-images.githubusercontent.com/10244644/171150234-29fe540e-090b-4bb8-8f4c-beb2a4c8c298.png">
3. Empty `span`.
<img width="525" alt="image" src="https://user-images.githubusercontent.com/10244644/171150509-31be45ff-6968-4cc2-90fd-285108a71fcc.png">
4. Complex code structure.

RabbiIslamRony:
@obiPlabon Issue has already been fixed https://github.com/sovware/directorist/pull/918

PR #916, opened 2022-05-31
Fix archive card with custom fields issue

obiPlabon:
#915


kowsar89:
Closed in the favor of #918

PR #927, opened 2022-06-12
Flush rewrite rules on install and update

obiPlabon:


vairafiq (commented):
seems it's not flushing rewrites

PR #934, opened 2022-06-15
files updated

vairafiq:


tanjimhasan (commented):
All changes are ok. But need to investigate why those changes are come? I think webpack generates /.min file in your machine in different way.

obiPlabon (approved):

PR #939, opened 2022-06-16
Radius search fix

kowsar89:


includes/model/SearchForm.php
```diff
@@ -328,20 +328,20 @@ public function buttons_template() {
 		}
 	}
 
-	public function range_slider_data( $data ) {
```
obiPlabon, line 331: Please do not delete any method without prior notice. We can deprecate first then after 3/4 releases we can hard deprecate or remove that method.

obiPlabon (changes requested):

includes/model/SearchForm.php
kowsar89: This method still exists in line 340

obiPlabon: Sorry, my bad.

vairafiq (approved):
Reviewed

vairafiq:
Searched radius is not checked in advanced filter

vairafiq:
Still not working https://www.awesomescreenshot.com/video/9556385?key=cae85b031c32738af28266bd7cb9508b

vairafiq:
Working perfectly now

obiPlabon (approved):
Working fine

PR #941, opened 2022-06-16
fix wrong redirection issue when needed to redirect to checkout + som…

M4hd1BD:
…e other stuff i don't know about


vairafiq (approved):
Reviewed

PR #949, opened 2022-06-20
Radius search URL issue fix

tanjimhasan:


assets/src/js/public/search-form.js
```diff
@@ -837,5 +838,27 @@ import './../global/components/select2-custom-control';
...
+            $('.directorist-location-js').each((index,locationDom)=>{
+                if($(locationDom).val() === ''){
+                    $(locationDom).closest('.directorist-advanced-filter').find('.directorist-advanced-filter__advanced--radius_search').css({display: "none"});
+                }else{
+                    $(locationDom).closest('.directorist-advanced-filter').find('.directorist-advanced-filter__advanced--radius_search').css({display: "block"})
+                    directorist_callingSlider();
+                }
+            });
+        }
+        $('body').on('keyup keydown input change focus', '.directorist-location-js', function (e) {
```
obiPlabon, line 852: only the `input` event is enough here. and maybe we can use debounce to improve performance.

obiPlabon: Doesn't work as expected https://www.awesomescreenshot.com/video/9556385?key=cae85b031c32738af28266bd7cb9508b

PR #977, opened 2022-06-29
Fix/double dom

tanjimhasan:


kowsar89:
### Issue Details
Search Home
Double Dom Element after ajax success
https://prnt.sc/FklXA8MNOH9J

templates/search-form-contents.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
kowsar89, line 2: unnecessary spacing change

templates/search-form-contents.php
```diff
@@ -7,40 +8,44 @@
 
 use \Directorist\Helper;
 
-if ( ! defined( 'ABSPATH' ) ) exit;
+if (!defined('ABSPATH')) exit;
```
kowsar89, line 11: unnecessary spacing change

kowsar89 (changes requested):
spacing updated in many places which is unnecessary

templates/search-form-contents.php
tanjimhasan: Space removed

kowsar89 (approved):
works fine

PR #981, opened 2022-06-30
Fixed php warning in search form

kowsar89:
### Issue details
Search Form
PHP warning when zip code is present but radius search is not present
http://prntscr.com/6v2mnGmKb-Lm


vairafiq (commented):
image

Better not to remove the property forever.

PR #985, opened 2022-07-03
fixed map-view js and range slider js error

devShamim:
map view issue => https://prnt.sc/pPG3PtqAVgM3
range slider issue => https://prnt.sc/SC-9-AnB8lsT


vairafiq (approved):
checked

PR #989, opened 2022-07-03
Price range bug fixed

kowsar89:
### Listing Form
Price Range not saving data if pricing type selected to only price range


vairafiq (approved):
looks good

PR #996, opened 2022-07-17
add hook to featured listings args

M4hd1BD:


vairafiq:
please use `directorst_` instead of `atbdp_`

vairafiq (changes requested):
please rename to `directorist_widget_featured_listings_query_arguments`

vairafiq (approved):
Looks good

PR #1014, opened 2022-07-18
Fixed advanced filter issue

devShamim:


kowsar89 (changes requested):
works fine, but files not built with correct node version

Issue #1046, opened 2022-07-25
Map not loading in single page map widget


kowsar89:
https://prnt.sc/enx4sGPDHaqz
How to regenerate: in single page, dont use location field from page builder

Ref: https://tasks.hubstaff.com/app/organizations/37274/projects/265387/tasks/4450197

kowsar89:
Fixed for openstreet map, still exists for google map

PR #1049, opened 2022-07-25
Fixed single listing widget map loading issue

devShamim:
https://github.com/sovware/directorist/issues/1046


kowsar89 (approved):
works fine

PR #1076, opened 2022-07-31
Security issues of templates dir

kowsar89:
Few escaping errors pending, will be fixed in upcoming commits

Note: directorist_submission_field_module filter has been removed, which is used in both pricing plan extensions


vairafiq (approved):
pass

PR #1077, opened 2022-08-01
Fix classes security issues

obiPlabon:
- `class-add-listing` - partial
- `class-ajax-handler` - partial
- `class-announcement` - done
- `class-background-updater` - done
- `class-cache-helper` - done
- `class-cron` - done
- `class-custom-post` - done
- `class-custom-taxonomy` - partial


vairafiq (approved):
pass

PR #1107, opened 2022-08-07
improve grammer, maybe

M4hd1BD:


vairafiq:
Need to change the template version according to assigned milestone

vairafiq (approved):
checked

PR #1117, opened 2022-08-09
deprecate/remove unused codes

kowsar89:
resolves id/22


vairafiq:
This function is required for renewal #1111 fixed the issue

PR #1131, opened 2022-08-17
login nonce conflict when registration form use in the same page

vairafiq:
user login shortcode in registration page to reproduce the issue


kowsar89:
Unable to reproduce

kowsar89 (approved):
seems working

PR #1137, opened 2022-08-23
remove file limit notice from custom field file upload

vairafiq:
image


kowsar89 (changes requested):
what about these lines? maybe remove them too if not used? https://prnt.sc/mYfZoSNSiWil

vairafiq:
We need to work on file upload template soon and removing input field might cause js error or increase the complexity of quick release(requested by theme and support team). So we may skip it for now and just remove the notice.

kowsar89 (approved):
looks good

PR #1139, opened 2022-08-23
Fix/form select

vairafiq:


kowsar89 (changes requested):
please don't change formatting unless necessary

kowsar89:
#1142

PR #1141, opened 2022-08-23
get rid of outdated code that assinged compard based on fields type

M4hd1BD:


kowsar89 (commented):
What problem does it solves?

kowsar89 (approved):

PR #1144, opened 2022-08-24
Refactor taxonomy utilities

obiPlabon:
Instead of using `wp_get_object_terms` we are going to use `get_the_terms` which usages cache.


adminazhar (commented):
@obiPlabon LGTM. It would save a lot of queries by caching.

One suggestion: I prefer to squash all the commits before moving to review(by rebasing), i know it can be done while merging, but its a good practice to have a single commit for small changes.

obiPlabon:
> One suggestion: I prefer to squash all the commits before moving to review(by rebasing), i know it can be done while merging, but its a good practice to have a single commit for small changes.

@adminazhar Thank you for the suggestion

vairafiq (approved):

PR #1168, opened 2022-09-06
Small fix for global posts leading to "directorist_author_profile" shortcut-page

AbyssWaIker:
In our theme (typer) url to author profile is taken not from login but from "user_nicename".

So everything I needed to do is to add is get_user_by( 'slug', $user_id) in addition to existing get_user_by( 'login', $user_id).

But if "get user_by()" doesn't find matching user (with \WP_User::get_data_by( $field, $user_id )) it just returns current user (making "get_current_user_id()" in the function redundant).

So I made a small refactor to make code cleaner and add slug check


AbyssWaIker:
Turned out my entire use-case was based on the fact that "regenerate page" has overriden "front user profile" page...
It's not really needed anymore.

Still, this little commit makes code a bit more readable, maintanable and prepared for links in wp's "user_nicename" format (specifically made to be url-friendly).

PR #1184, opened 2022-09-09
Fix for pagination in multilingual (wpml) directory

AbyssWaIker:
In our wpml setup we have languages in separate "directories" (like "site", "site/pt-pr", "site/pt-br', "site/ru", "site/uk"). 

So when (at line 62) we attempt to remove home_url ("site/uk") from link ("site/permalink") it doesn't find the string and leaves the whole site address inside "pages" rewrite rule. 

Leading to 404 when we our non english speaking visitors are trying to see page 2 of directory.


vairafiq:
Hi @AbyssWaIker,

Thanks for your contribution. As we have an integration plugin for WPML, this code needs to be placed in [this](https://wordpress.org/plugins/directorist-wpml-integration/) plugin. And we are going to add a hook in the core plugin.

PR #1201, opened 2022-09-20
add translation support for search page title

M4hd1BD:


vairafiq (commented):
image
We can avoid this

Issue #1199, opened 2022-09-19
string non translatable


cebas:
I created #1126 that was closed, but the bug is still there. So I open a new one.
Check my comment there: 
_Originally posted by @cebas in https://github.com/sovware/directorist/issues/1126#issuecomment-1245593887_

PR #1204, opened 2022-09-22
fix placeholder for price range and price unit field

M4hd1BD:


M4hd1BD:
Also fixes #1199

kowsar89 (changes requested):
Please change template version

kowsar89 (approved):
Looks good

PR #1219, opened 2022-10-06
Added a new default image size option for all-listings page

kowsar89:
Screenshot: https://prnt.sc/qmhvjKmF9K-R

It'll improve page load time on all-listings page. Improvement example: https://prnt.sc/CnmppY2UO5R2

### DIFF
--- a/includes/classes/class-settings-panel.php
+++ b/includes/classes/class-settings-panel.php
@@ -1849,7 +1849,7 @@ Please remember that your order may be canceled if you do not make your payment 
                     'options' => [
                         [
                             'value' => 'directorist_preview',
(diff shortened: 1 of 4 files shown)


includes/classes/class-settings-panel.php
```diff
@@ -1844,6 +1847,10 @@ public function prepare_settings()
                     'type'  => 'select',
                     'value' => 'large',
                     'options' => [
+                        [
+                            'value' => 'directorist_preview',
+                            'label' => sprintf( __( 'Default', 'directorist' ), $default_preview_size_text ),
```
obiPlabon, line 1852: sprintf doesn't have any placeholder here.

obiPlabon (approved):

PR #1220, opened 2022-10-06
add error check in accessing directory form fields

vairafiq:
Issue:  https://tasks.hubstaff.com/app/organizations/37274/projects/265387/tasks/4768877


kowsar89 (approved):
seems ok

PR #1222, opened 2022-10-10
add additional check for outdated extensions

vairafiq:
Issue: https://tasks.hubstaff.com/app/organizations/37274/projects/302330/tasks/4728657


kowsar89 (approved):
seems ok

PR #1239, opened 2022-10-25
fix an issue with wrong operator getting assigned if there is multipl…

M4hd1BD:
…e custom fields

### DIFF
--- a/includes/model/Listings.php
+++ b/includes/model/Listings.php
@@ -875,7 +875,7 @@ class Directorist_Listings {
 					}
 				} else {
 					$field_type = str_replace( 'custom-', '', $key );
-					$field_type = preg_replace('/([!^0-9])|(-)/', '', $field_type);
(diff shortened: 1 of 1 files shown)


includes/model/Listings.php
```diff
@@ -875,6 +875,7 @@ public function parse_search_query_args() {
 					}
 				} else {
 					$field_type = str_replace( 'custom-', '', $key );
+					$field_type = preg_replace('/([!^0-9])|(-)/', '', $field_type);
```
kowsar89, line 878: 1. Use inline comment explaining what this regular expression does
2. Follow PHP coding standards for space usage: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#space-usage

kowsar89 (changes requested):

includes/model/Listings.php
M4hd1BD: Done

kowsar89 (approved):

PR #1245, opened 2022-11-01
instant search now use options from all listings

### LINKED ISSUE #1233: Sort by title and listing per page breaks on pagination with Instant Search
### Directorist version

7.4.2(alpha)

### Environment

_No response_

### Description

Paginated pages don't respect sort by title, it shows listing randomly, it also doesn't respect the listing per page settings.

### Steps to reproduce

1. Settings > All Listing - Set Orderby Title and Sory By Ascending and also set Listing Per Page to 2. Also turn on Instant Search
2. Go to All Listings page, you'll see 2 listings, sorted by their title, lets say they start with A and B respectively
3. Now go to the Second page using the Paginator, you would expect it to show listings that start with C and D, but instead you will see more than 2 listings that doesn't follow the title sorting convention. 

### Isolating the problem

- [X] This bug happens with a default WordPress theme active (eg. 2021, 2022 etc).

kowsar89:
Instant search now use options from all listings instead of search result
Fixes #1233

### DIFF
--- a/includes/classes/class-ajax-handler.php
+++ b/includes/classes/class-ajax-handler.php
@@ -135,7 +135,7 @@ if ( ! class_exists( 'ATBDP_Ajax_Handler' ) ) :
 				$_POST['ids']    = $args['ids'];
 			}
 
-			$listings = new Directorist\Directorist_Listings( $args, 'search_result' );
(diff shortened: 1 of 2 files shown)


vairafiq (approved):
checked

PR #1269, opened 2022-11-20
Fix nonce securities

obiPlabon:
## PR Type
- [x] Bugfix
- [x] Security fix
- [x] Refactoring (no functional changes, no API changes)

## Description
Please check the security group message for more details...

Note: Refactored ```login.php``` template, it's a huge template with legacy codes (looks like)... So, I thought let's reformat. Please check this template carefully.

## Checklist


kowsar89 (changes requested):
Right now the vulnerable function "send_system_info" isn't used anywhere, hence the existence of full codebase is unnecessary.

This function is called only in here which is unused as well since it's "display:none" always, hence should be removed: https://prnt.sc/AMXUXIg3ZORH

History: Sending system info feature was removed at later point, but it seems like legacy code wasn't removed.

includes/system-status/custom-url.php
```diff
@@ -7,61 +7,77 @@ class ATBDP_Custom_Url
...
 
-		if ( ! isset( $_GET['atbdp-system-info'] ) || empty( $_GET['atbdp-system-info'] ) ) {
+    public function view_debug_info() {
+		if ( empty( $_GET['directorist_debug'] ) ) {
 			return;
 		}
 
-		$queryValue = sanitize_text_field( wp_unslash( $_GET['atbdp-system-info'] ) );
-		$token      = get_transient( 'system_info_remote_token' );
+		if ( ! directorist_verify_nonce( 'directorist_debug_nonce', 'directorist_get_debug_info' ) ) {
```
kowsar89, line 59: Unable to visit remote viewing URL as guest: https://prnt.sc/_txlO6EvbExo
Nonce is blocking this.

kowsar89 (changes requested):

includes/system-status/custom-url.php
```diff
@@ -7,61 +7,77 @@ class ATBDP_Custom_Url
...
+		}
+
+		if ( ! current_user_can( 'manage_options' ) ) {
+			wp_send_json_error( __( 'You are not allowed to create secret url', 'directorist' ),  403 );
+		}
+
 		$token   = wp_rand();
 		$expires = apply_filters( 'atbdp_system_info_remote_token_expire', DAY_IN_SECONDS * 3 );
+
 		set_transient( 'system_info_remote_token', $token, $expires );
```
kowsar89, line 25: Alternative: Maybe instead of transient, we can use wp_options. There's no major benefit for using transient in here.

templates/account/login.php
```diff
@@ -13,207 +13,215 @@
...
+									$message .= '<a href="' . esc_url( add_query_arg( $link, ATBDP_Permalink::get_login_page_url() ) ) . '">' . esc_url( add_query_arg( $link, ATBDP_Permalink::get_login_page_url() ) ) . '</a>';
+
+									$message = atbdp_email_html( $subject, $message );
+
+									$headers[] = 'Content-Type: text/html; charset=UTF-8';
+									$mail      = wp_mail( $email, $subject, $message, $headers );
+									if ( $mail ) {
+										$success = __( 'A password reset email has been sent to the email address on file for your account, but may take several minutes to show up in your inbox.', 'directorist' );
+									} else {
+										$error = __( 'Password updated! But something went wrong sending email.', 'directorist' );
```
kowsar89, line 190: Suggestion: misleading text. Could be like: "Something went wrong, unable to send the password reset email. If the issue persists please contact with the site administrator."

kowsar89 (changes requested):
template version need updating, probably 7.4.3.1

includes/system-status/custom-url.php
obiPlabon: Transient will be cleared automatically but the option won't. I think that's why it's being used here.

includes/system-status/custom-url.php
obiPlabon: Nonce won't work, so removed.

kowsar89 (changes requested):
Have some PHPCS issues

kowsar89 (changes requested):
After generating URL, view button doesn't comes up: https://prnt.sc/k1Sw0QRw5iZ-

Reproduction steps:
- Revoke URL
- Reload
- Create URL
- View button missing

kowsar89 (dismissed):
LG

kowsar89 (changes requested):
Frontpage stops loading with notice: https://prnt.sc/pWp-SQzZ__8y

kowsar89 (approved):
Looks good

PR #1279, opened 2022-11-27
Fix double alert issue on require login notice

### LINKED ISSUE #1268: Unwanted code in account.js, conflicting with single-listing.js
### Directorist version

latest

### Environment

any

### Description

While using the account.js script for popup login and registration. the single listing report button "login alert" shows twice. 

I checked and understand that some code is duplicated in the account.js 
**path** directorist/assets/js/account.js
**line number** 208-212

### Steps to reproduce

To reproduce the issue, please use the OneListing Theme.

### Isolating the problem

- [ ] This bug happens with a default WordPress theme active (eg. 2021, 2022 etc).

devShamim:
## PR Type

- [x] Bugfix

## linked issue
Fixes #1268

### DIFF
--- a/assets/src/js/public/components/login.js
+++ b/assets/src/js/public/components/login.js
@@ -41,8 +41,6 @@
         $('.atbdp-require-login, .directorist-action-report-not-loggedin').one('click', function (e) {
             e.preventDefault();
             alert(directorist.login_alert_message);
-            $('.directorist-report-abuse-modal').css('display', 'none');
-            throw {message : directorist.login_alert_message};
         });
     });
 })(jQuery);
--- a/assets/src/js/public/components/login.js
+++ b/assets/src/js/public/components/login.js
@@ -38,16 +38,11 @@
         });
 
         // Alert users to login (only if applicable)
-        $('.atbdp-require-login, .directorist-action-report-not-loggedin').on('click', function (e) {
+        $('.atbdp-require-login, .directorist-action-report-not-loggedin').one('click', function (e) {
             e.preventDefault();
-            var loginAlert = document.createElement("div");
-            loginAlert.setAttribute("style","position:fixed;top:30px;left:50%;background-color:#2f2f2f;z-index:9999;transform:translateX(-50%);padding:15px;text-align:center;color:#fff;border-radius:4px;");
-            loginAlert.innerHTML = directorist.login_alert_message;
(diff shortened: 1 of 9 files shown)


kowsar89 (changes requested):
behavior changed

kowsar89 (changes requested):
Issue isn't fixed in onelisting theme

assets/src/js/public/components/login.js
```diff
@@ -38,10 +38,18 @@
         });
 
         // Alert users to login (only if applicable)
-        $('.atbdp-require-login, .directorist-action-report-not-loggedin').on('click', function (e) {
+        $('.directorist-action-report-not-loggedin').on('click', function (e) {
             e.preventDefault();
             alert(directorist.login_alert_message);
-            return false;
+            $('.directorist-report-abuse-modal').length ? $('.directorist-report-abuse-modal').css('display', 'none') : '';
+            throw {message : directorist.login_alert_message};
```
kowsar89, line 45: console error: http://prntscr.com/rmJPO0ecepid

assets/src/js/public/components/login.js
```diff
@@ -38,10 +38,18 @@
         });
 
         // Alert users to login (only if applicable)
-        $('.atbdp-require-login, .directorist-action-report-not-loggedin').on('click', function (e) {
+        $('.directorist-action-report-not-loggedin').on('click', function (e) {
             e.preventDefault();
             alert(directorist.login_alert_message);
-            return false;
+            $('.directorist-report-abuse-modal').length ? $('.directorist-report-abuse-modal').css('display', 'none') : '';
```
kowsar89, line 44: is this really required?

kowsar89 (changes requested):

assets/src/js/public/components/login.js
devShamim: yes, it's required

devShamim: > is this really required?

yes, it's required

assets/src/js/public/components/login.js
devShamim: > console error: http://prntscr.com/rmJPO0ecepid

I had to throw an error to stop the event bubbling since stopPropagation can't use here.

Issue #1229, opened 2022-10-18
optimize get_users call on class-settings-panel and class-announcement 


M4hd1BD:
### Directorist version

7.4.1

### Description

this is not ideal for sites with large number of users, the server will run out of memory and it won't be possible to access the settings panel.

### Steps to reproduce

1. Have a site with a large number of users (over 30k or maybe less, depending on your server)
2. Go to Settings and you'll see its just returning blank page. If you see the logs you'll see memory out error. 

### Isolating the problem

- [X] I have reproduced this bug in localhost.
- [X] This bug happens with a default WordPress theme active (eg. 2021, 2022 etc).

M4hd1BD:
Temporary fix:
Limit the call to load 1k users only: 
![image](https://user-images.githubusercontent.com/37076749/196609694-e62ce76d-2de8-453e-929d-343331533289.png)
![image](https://user-images.githubusercontent.com/37076749/196609865-d015a767-7c1b-4f3b-9200-9802cae68226.png)

kowsar89:
From https://10up.github.io/Engineering-Best-Practices/php/

Do not use posts_per_page => -1.

This is a performance hazard. What if we have 100,000 posts? This could crash the site. If you are writing a widget, for example, and just want to grab all of a custom post type, determine a reasonable upper limit for your situation.
```
<?php
// Query for 500 posts.
new WP_Query( array(
  'posts_per_page' => 500,
));
?>
```

kowsar89:
> Temporary fix: Limit the call to load 1k users only: ![image](https://user-images.githubusercontent.com/37076749/196609694-e62ce76d-2de8-453e-929d-343331533289.png) ![image](https://user-images.githubusercontent.com/37076749/196609865-d015a767-7c1b-4f3b-9200-9802cae68226.png)

In that case, all users will not get the announcements

PR #1282, opened 2022-11-30
Filter hook added in announcement user query

kowsar89:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
Limit is 1000 now.

**Side effects for existing users:** Now announcement notice/emails will not be sent to more than 1000 users unless user modify the hook.

To change this as unlimited using hook:

`
add_filter( 'directorist_announcement_user_query_num', function() {
	return -1;
} );
`

## Any linked issues
Temporary solution for #1229

## Checklist

### DIFF
--- a/includes/classes/class-announcement.php
+++ b/includes/classes/class-announcement.php
@@ -403,7 +403,7 @@ if ( ! class_exists( 'ATBDP_Announcement' ) ) :
 					array(
 						'role__not_in' => 'Administrator',   // Administrator | Subscriber
 						'fields'       => 'user_email',
-						'number'       => apply_filters( 'directorist_announcement_user_query_num', '' ),
+						'number'       => apply_filters( 'directorist_announcement_user_query_num', 1000 ),
 					)
 				);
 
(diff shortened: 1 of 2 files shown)


includes/classes/class-announcement.php
```diff
@@ -403,6 +403,7 @@ public function send_announcement() {
 					array(
 						'role__not_in' => 'Administrator',   // Administrator | Subscriber
 						'fields'       => 'user_email',
+						'number'       => apply_filters( 'directorist_announcement_user_query_num', '' ),
```
vairafiq, line 406: isn't it better to cover the total arguments instead of the number only?

kowsar89: I'm using the same filter in two places. If we cover the whole arguments:

- we have to use 2 different filters instead of 1, because arguments are different
- it'll be a tough decision if we decide to remove this filter later, because more arguments will depend on it
- Since we have plans to improve this query later, we may refactor the whole query(and maybe remove the filter). In that case, covering whole arguments might make things complicated

vairafiq: OK, looks like the default value is missing

kowsar89: default value is empty string: https://github.com/WordPress/wordpress-develop/blob/9690aa7f7b0b7acd5acad54278952622585dec16/src/wp-includes/class-wp-user-query.php#L110

vairafiq (approved):
looks good

PR #1304, opened 2022-12-15
Fixed Adding More Than One Tag Issue

### LINKED ISSUE #1303: can't add more than one new tag
### Directorist version

v7.4.6

### Environment

_No response_

### Description

if allow new is turned on for the tag field, user can't add more than one new tag using the add listing form

### Steps to reproduce

https://www.loom.com/share/031d9a7d0db2459c98e0e5e752b909c0

### Isolating the problem

- [X] This bug happens with a default WordPress theme active (eg. 2021, 2022 etc).

iamibrahimriaz:
## PR Type

- [x] Bugfix

## Any linked issues
Fixes #1303

## Checklist


vairafiq:
It's working now. The location field has the same issue which has also been fixed with the same code. Thanks @iamibrahimriaz 

I am going to merge it on @devShamim review

devShamim (changes requested):
Please use specified node version in package.json file and regenerate the codes again.
image

PR #1364, opened 2023-02-15
Fix listing meta empty field saving from backend

obiPlabon:
## PR Type

- Bugfix
- Improvement
- Refactoring (no functional changes, no API changes)

## Description
Whenever a listing is created or updated from the backend editor, all the empty fields were also being saved. Now only the non-empty field will be saved.
Before:
before-refactoring
After:
after-refactoring

## Checklist


Armanul46:
Works fine

vairafiq (approved):

PR #1397, opened 2023-03-08
fixed directory builder widget label issue

syedgalib:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
The card option fields in the 'card-builder' widget uses label from its root, but if the option fields contains label field in it then it uses that value instead of the root label.

The back button widget has a toggle field and its key is defined as 'label' which is not appropriate for its kind.
Therefore the card option field was using its boolean value as the widget label.

So I have added an additional check on the 'getWidgetLabel' helper to skip the label field if its type is not text.

## Any linked issues
[Task: Fix - Directory Builder - Listing Header - Back Button Issue](https://tasks.hubstaff.com/app/organizations/37274/projects/265386/tasks/5316474)

## Checklist

### DIFF
--- a/assets/src/js/admin/vue/modules/form-fields/Card_Builder_Listing_Header_Field.vue
+++ b/assets/src/js/admin/vue/modules/form-fields/Card_Builder_Listing_Header_Field.vue
@@ -6,7 +6,7 @@
         v-bind="widgetCardOptionsWindow"
         @close="closeCardWidgetOptionsWindow()"
       />
-      
+
       <options-window
         :active="widgetOptionsWindowActiveStatus"
         v-bind="widgetOptionsWindow"
@@ -200,7 +200,7 @@ export default {
     output_data() {
       let output = {};
       let layout = this.local_layout;
-      
+
       // Parse Layout
       for ( let section in layout ) {
         output[section] = {};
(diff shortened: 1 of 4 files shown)


assets/src/js/admin/vue/modules/form-fields/Card_Builder_Listing_Header_Field.vue
```diff
@@ -772,10 +772,11 @@ export default {
         label = widget.label;
       }
 
-      if ( 
-        this.isObject( widget.options ) && 
-        widget.options.fields && 
-        widget.options.fields.label && 
+      if (
```
obiPlabon, line 775: How about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

`widget?.options?.fields?.label?.value?.length && widget?.options?.fields?.type === 'text'`

obiPlabon (changes requested):

assets/src/js/admin/vue/modules/form-fields/Card_Builder_Listing_Header_Field.vue
syedgalib: Sure

obiPlabon (approved):
Maybe ok

PR #1404, opened 2023-03-14
improve some code for PHP8.1

mdalaminbey:
1. set `get_directorist_option` force default
2. rename appsero client `__trans` to `_trans`
3. set DateTime methods return type
4. create `sniffer:php8.1` composer command


obiPlabon (approved):
Looks good

PR #1564, opened 2023-08-16
Refactor review system

vairafiq:
1. Add new hooks
2. Add new parameters in existing hooks
3. Add review fraction star
4. Tweaking review calculation


obiPlabon (changes requested):
PR title says "add new hook and parameter" but actually there's more than that. Please explain other changes.

PR #1594, opened 2023-09-13
Improve/builder data

syedgalib:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
The directory builder is JSON based builder and its data is provided from the `Multi_Directory_Manager` class. But for some reason  if any other class or function requires only the builder's JSON data, the entire class will have to be executed which is not necessary.

Thats why I have extracted the builder JSON data into a separate class so that the builder's JSON data can be access from any where without executing the main class.

## Checklist


obiPlabon (changes requested):
- We don't need a data sub-directory. Please keep it simple and under the `multi-directory-setup`
- No need for `get_data_content` we know the file exists. So no need to check and require, just keep it the way as it was before.
- Rename `builder-data.php` to `class-builder-data.php`
- Instead of returning all the data as an array better return separately, like `get_fields`, `get_lyouts`, `get_config`, `get_options`.
- Since `prepare_settings` has been called 6 times we can cache the assignment, like the following -
```php
public function prepare_settings() {
  if ( empty( self::$fields ) ) {
	  self::$fields  = Builder_Data::get_fields();
  }
  
  if ( empty( self::$layouts ) ) {
	  self::$layouts = Builder_Data::get_layouts();
  }
  
  if ( empty( self::$config ) ) {
	  self::$config  = Builder_Data::get_config();
  }
        
  if ( empty( self::$options ) ) {
	  self::$options = Builder_Data::get_options();
  }
}
```

obiPlabon (approved):
Looks good

PR #1619, opened 2023-10-05
Fix review submission captcha issue

obiPlabon:
## PR Type

- Bugfix

## Description
Since the Directorist is using WordPress's default comment system to store reviews. So, whenever there is a captcha or any other captcha-like system it doesn't users to submit any review and usually returns a 403 error.

Note: only the source files have been changed, to test the PR someone must run `npm run dev` or `npm run prod`.

## Checklist


vairafiq (approved):
Looks good

PR #1621, opened 2023-10-08
Fixed text that are not translatable but run through translation functions

### LINKED ISSUE #1311: Dynamic strings (like page title) being translated make translations grow without limit
Currently, any search made by a visitor is passed to __() function. Also, page titles of listings seems to be passed to __().

This behavior is found on **includes/classes/class-seo.php**

This cause that translation plugins that store translation strings on a database, like TranslatePress, keep growing its tables just to store anything that is searched on the site or dynamic page titles that do not need to be translated.

Listings titles are mainly names, like a restaurant name, person name, etc. It should not be translatable the whole string. If such functionality is desired, that could be achieved using variables in the translation string (to avoid storing repeated strings).

obiPlabon:
## PR Type

- Bugfix
- Improvement

## Any linked issues
Fixes #1311 

## Checklist


vairafiq (approved):
Looks good

PR #1625, opened 2023-10-11
Refactor never expire listings query

obiPlabon:
## PR Type

- Improvement

## Description
Refactored all the `_never_expire` meta query, so that it only queries by key, not by value. 

## Checklist


vairafiq (approved):
Looks good

PR #1643, opened 2023-11-07
Refactor gettype function call

obiPlabon:
## PR Type

- Improvement

## Description
Refactored all `gettype()` function call with respective type checking call.
For example, `'array' === gettype( $value )` with `is_array( $value )` and so on...

## Checklist


vairafiq (approved):
looks good

PR #1669, opened 2023-11-30
Fix fields options validation

obiPlabon:
## PR Type

- Bugfix

## Description
If a select, radio, and checkbox field value contains < (less than) character then it becomes encoded ($lt;) when saved. And because of this, the field value validation doesn't return the expected result.

## Checklist


vairafiq (approved):
looks good

PR #1678, opened 2023-12-31
Fixed empty image field undefined index issue

obiPlabon:
## PR Type

- Bugfix

## Description
When the image field is required but no image is selected then it not only shows the required message but also throws an undefined index error.

## Any linked issues
Fixes #

## Checklist


vairafiq (approved):
works fine

PR #1715, opened 2024-03-19
Improve price field

obiPlabon:
## PR Type

- Bugfix
- Improvement

## Description
When price type is selected as range or unit then in some of our themes the add listing form couldn't save the price range or unit value due to the missing price type value.

Now the system is backward compatible and doesn't require any update from any side.

## Any linked issues
Fixes # internal

## Checklist

### DIFF
--- a/assets/js/account.js
+++ b/assets/js/account.js
@@ -94,7 +94,6 @@
 /***/ (function(module, exports) {
 
 ;
-
 (function ($) {
   window.addEventListener('DOMContentLoaded', function () {
     // Recovery Password Modal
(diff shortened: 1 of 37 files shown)


Armanul46:
https://www.loom.com/share/ccfafdc86d2d42c8a7bc871f43e938d2?sid=7adf5668-88e4-4c59-bbe6-c43515e53159

vairafiq:
Two different experiences in admin and frontend forms.

1. Frontend: Saves only selected type and reset the deselected type
2. Backend: Saves both type in form submission

vairafiq (approved):
The form-saving experience has been fixed in another PR. Everything else looks good

PR #1743, opened 2024-05-12
Improved mono directory listing queries

obiPlabon:
## PR Type

- Improvement

## Description
Usually, users build a single site. So, when multi-directory is disabled there's no need to include the directory query along with the listings query. And that's why we made the directory query conditional to improve the query performance of a single directory site.

## Any linked issues
Fixes #

## Checklist


vairafiq (commented):
If this option is disabled for temporary purposes and he wants to enable multi-directory later, there will be some potential issues:

1. Show all directory listings under the default directory.
2. If any user updates a listing from any other directory, the directory type will be changed to the default directory.

PR #1862, opened 2024-09-09
Instant search

Armanul46:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description

1. left/right sidebar instant search conflict with no sidebar instant search
2.
3.

## Any linked issues
Fixes #

## Checklist

### DIFF
--- a/assets/src/js/public/components/instantSearch.js
+++ b/assets/src/js/public/components/instantSearch.js
@@ -1103,7 +1103,9 @@ import debounce from '../../global/components/debounce';
             },
             success: function (html) {
                 if (html.search_result) {
-                    $(_this).closest('.directorist-instant-search').find('.directorist-header-found-title').replaceWith(html.header_title);
+                    $(_this).closest('.directorist-instant-search').find('.directorist-header-found-title').remove();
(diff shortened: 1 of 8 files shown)


assets/src/js/public/components/instantSearch.js
```diff
@@ -1143,17 +1145,17 @@ import debounce from '../../global/components/debounce';
     }
 
     // sidebar on keyup searching
-    $('body').on("keyup", ".directorist-instant-search form", debounce( function(e) {
+    $('body').on("keyup", ".directorist-instant-search .listing-with-sidebar form", debounce( function(e) {
```
vairafiq, line 1148: The `listing-with-sidebar` class, when used with a `keyup` event, may cause conflicts with other themes or plugins. Prefixing is required to avoid this issue.

vairafiq (changes requested):

assets/src/js/public/components/instantSearch.js
Armanul46: There is no chance of conflict because we always use the parent div class, like .directorist-instant-search, for listing-with-sidebar.

PR #2161, opened 2024-12-07
Importer hooks

HeyMehedi:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
Added hooks for Gallery Image importing,

## Any linked issues
Fixes #

## Checklist


Armanul46:
Looks good

vairafiq (changes requested):
1. Need to add hook doc block with @since and other necessary
2. Need to pass the current loop item so that anyone can process data based on the loop. Otherwise, the hooks are quite useless

includes/classes/class-tools.php
```diff
@@ -109,6 +109,16 @@ public function atbdp_import_listing() {
...
 
+            /**
+			 * Fires before the listings import process starts.
+			 *
+			 * This action allows developers to execute custom tasks or add integrations
+			 * Before the listing import process begins.
+			 *
+			 * @since 8.0.11
+			 */
+			do_action( 'directorist_before_listings_import' );
```
vairafiq: Better to pass all the request data `do_action('directorist_before_listings_import', $_POST);`

vairafiq (changes requested):

includes/classes/class-tools.php
vairafiq: do the same for both before after hooks

HeyMehedi:
> 2. Need to pass the current loop item so that anyone can process data based on the loop. Otherwise, the hooks are quite useless

Thank you for your feedback! I see your point about passing the current loop item to hooks, but I’d like to clarify that there are already hooks placed inside the loop for this purpose. Specifically, the _directorist_import_listing_starts_ and _directorist_listing_imported_ hooks allow developers to process individual loop items effectively during the import process.

The hooks you mentioned—_directorist_before_listings_import_ and _directorist_after_listings_import_—are intentionally placed outside the loop. They serve a different purpose, such as initializing or finalizing tasks related to the entire import process, rather than processing individual items.

If you’re looking for something specific that isn’t addressed by the current hooks, feel free to share more details. I’d be happy to explore further improvements. Thanks for raising this point! 😊

includes/classes/class-tools.php
HeyMehedi: Since `$_POST` is processed before the loop, I have passed the sanitized `$posts` instead of directly using the raw `$_POST`. Additionally, I have included `$data` in the last hook. I believe this approach should be good to go.

vairafiq (approved):

PR #2184, opened 2024-12-29
Field Specific Query Fix

HeyMehedi:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
The search functionality was returning zero results when a custom field contained a '-'. This happened because the '-' was being treated as a key-value pair. While this behavior was intended for numeric fields, the necessary check to restrict it to only a **number** fields, causing the issue. This update resolves it by ensuring the handling is limited to number fields.

## Taiga Issue 
No: 517

## Checklist


includes/model/Listings.php
```diff
@@ -930,7 +930,7 @@ public function parse_search_query_args() {
 					$field_type = str_replace( 'custom-', '', $key );
 					$field_type = preg_replace( '/([!^0-9])|(-)/', '', $field_type ); //replaces any additional numbering to just keep the field name, for example if previous line gives us "text-2", this line makes it "text"
 					// Check if $values contains a hyphen
-					if ( strpos( $values, '-' ) !== false ) {
+					if ( 'number' === $field_type && strpos( $values, '-' ) !== false ) {
```
vairafiq, line 933: `if ( 'number' === (string) $field_type && is_string( $values ) && strpos( $values, '-' ) !== false ) {`

HeyMehedi: As it's a strict type comparison, we don't have to add the type hinting..

vairafiq (approved):

PR #2198, opened 2025-01-15
Add users query limit to published listings owner only

obiPlabon:
## PR Type

- Improvement

## Description
Getting all users' data using the rest API is possible, but it is not a good practice. So in this PR, user query is limited to only users with published listings.

## Any linked issues
Fixes #

## Checklist


HeyMehedi (approved):
looks good

PR #2285, opened 2025-03-19
Feat view count dynamic update

obiPlabon:
## PR Type

- Bugfix
- New Feature

## Description
Due to page caching view count doesn't update. So an ajax request is sent on page load. The request will get the latest count and will cache it for 5 hours, and once invalidated it'll send the ajax request again.

## Any linked issues
Fixes #

## Checklist


assets/src/js/public/components/update-view-count.js
HeyMehedi, line 1: While sites don't have any caching system, it may not be required stored on localStorage. We can provide an option to enable/disable this feature. 

In the first load, a listing shows count 2, and when the Ajax version comes it goes to 0. please check this SS: https://prnt.sc/uOMR_VJArAcC

Armanul46:
looks good

Armanul46 (approved):

PR #2293, opened 2025-03-25
Fixed listing preview title preview filter too few arguments issue

obiPlabon:
## PR Type

- Bugfix

## Description
We are prefixing the listing title with "Preview:" instead of the default "Private:" on the listing preview screen using the "the_title" filter hook. But it seems in some cases the filter hook doesn't have all the required arguments hence the user will experience "too few arguments" php fatal error.

## Any linked issues
Fixes #

## Checklist


Armanul46:
Looks good

Armanul46 (approved):

PR #2328, opened 2025-05-04
Improve: Polylang integration

syedgalib:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [ ] Improvement
- [x] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
- Fix: (Polylang) Fixed listing preview URL does not redirect to correct language path when listing is submitted.

## Any linked issues
Fixes #

## Checklist


directorist-base.php
```diff
@@ -205,6 +212,13 @@ public static function instance()
 
 			self::$instance->includes();
 
+			// Beta plugin lookup
+			$plugin_data = get_plugin_data( plugin_dir_path( __FILE__ ) . 'directorist-base.php' );
```
obiPlabon: we can simply use since it's the base file. `get_plugin_data( __FILE__ );`

directorist-base.php
```diff
@@ -236,6 +250,7 @@ public static function instance()
 			// self::$instance->validator = new ATBDP_Validator;
 			// self::$instance->ATBDP_Single_Templates = new ATBDP_Single_Templates;
 			self::$instance->tools = new ATBDP_Tools();
+			self::$instance->multilingual = new Directorist_Multilingual();
```
obiPlabon: What is the use of storing the reference? It's to remove action, then just make the `init` method of `Directorist_Multilingual` static.

includes/classes/class-add-listing.php
```diff
@@ -141,6 +141,8 @@ public function atbdp_submit_listing() {
 
 				$posted_data = wp_unslash( $_POST );
 
+				do_action( 'directorist_before_processing_ajax_request', [ 'action' => 'add_listing_action', 'params' => $posted_data ] );
```
obiPlabon: Please rename the hook. Better use
```php
// Option 1
do_action( 'directorist_before_listing_submission', $posted_data );

// Option 2
do_action( 'directorist_ajax_before_listing_submission', $posted_data );
```
Action name can be extracted from the `$_POST`

includes/classes/class-add-listing.php
```diff
@@ -462,6 +464,10 @@ public function atbdp_submit_listing() {
 
 				$data['redirect_url'] = urlencode( $data['redirect_url'] );
 
+				$data = apply_filters( 'directorist_submitted_listing_ajax_request_response', $data );
```
obiPlabon: Please rename the hook. Better use
```php
apply_filters( 'directorist_listing_submission_response', $data );
// Or
apply_filters( 'directorist_ajax_listing_submission_response', $data );
```

includes/classes/class-add-listing.php
```diff
@@ -462,6 +464,10 @@ public function atbdp_submit_listing() {
...
+
+				do_action( 'directorist_after_processing_ajax_request', [ 'action' => 'add_listing_action', 'params' => $posted_data,  'response' => $data ] );
```
obiPlabon: Please rename the hook. Better use
```php
do_action( 'directorist_after_listing_submission', $posted_data );
// Or
do_action( 'directorist_ajax_after_listing_submission', $posted_data );
```
Action name can be extracted from the $_POST

obiPlabon (changes requested):

includes/classes/class-add-listing.php
syedgalib: The point of choosing the 'directorist_before_processing_ajax_request' hook name is to target any Ajax request from Directorist, not just the listing submission. Because the translation task for the Ajax request is common for any Ajax request. 

Currently the hook is placed only to the submission form, but later if any other Ajax request needs translation then it can be done by just by placing the common hook.

I suggest 'directorist_ajax_before_request_handling' for the new hook name, what do you think?

directorist-base.php
syedgalib: It's just to follow the existing patterns. Updated by the way.

directorist-base.php
syedgalib: Updated. It was previous code, just moved from one place to another.

includes/classes/class-add-listing.php
syedgalib: I suggest `directorist_ajax_after_request_handling`. What do you think?

obiPlabon (changes requested):
image
Builder doesn't load. Looks like there's a git conflict!

obiPlabon:
image

We are calling `$plugin_data = get_plugin_data( __FILE__ );` too early.

obiPlabon (approved):

PR #2378, opened 2025-05-28
Added filter hooks in listing form submission

MahfuzulAlam:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
Added a couple of hooks -

1. directorist_guest_user_account_creation
2. directorist_insert_listing_postdata

## Any linked issues
Fixes #

## Checklist


obiPlabon:
Guest users cannot submit listings without account creation, cause an author is required for a listing. So why do we need a filter hook when the guest submission can be disabled from the settings panel?

MahfuzulAlam:
Thank you for asking! There are a few exceptions where users have requested this feature: they want their visitors to be able to submit listings without needing to create an account. Essentially, they prefer an open form that anyone can use to post a listing on their site.
Since we don’t plan to introduce this kind of weird functionality as a built-in feature, I’ve provided them with a solution using hooks. This way, they can achieve what they need while we continue to follow our standard process.

includes/classes/class-add-listing.php
```diff
@@ -372,6 +387,22 @@ public function atbdp_submit_listing() {
...
+                     *
+                     * @since 8.4.4
+                     *
+                     * @param array $listing_data The array of post data to be inserted, compatible with wp_insert_post().
+                     * @param array $posted_data  The raw submitted form data from the listing submission form.
+                     * @param array $form_fields  The list of form fields used in the submission form.
+                     *
+                     * @return array Modified post data array to be inserted as a listing.
+                     */
+                    $listing_data = apply_filters( 'directorist_insert_listing_postdata', $listing_data, $posted_data, $form_fields );
```
obiPlabon: `directorist_insert_listing_data`

obiPlabon (changes requested):

obiPlabon (approved):

PR #2395, opened 2025-06-18
Formgent integration

syedgalib:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [ ] Improvement
- [x] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
- Added - API - Install and activate plugin.
- Added - FormGent widget for single listing content.

## Any linked issues
Fixes #

## Checklist


iamibrahimriaz:
Looks Good

obiPlabon (changes requested):
1. Instead of using (_) underscore use hyphen (-) in the template name. Yes, we used that before but don't wanna make the same mistake again.
2. When Formgent is disabled don't render the template content, instead return early.
3. Why don't we save the form ID instead of the shortcode?

obiPlabon (approved):

PR #2580, opened 2025-11-06
Fix: Search does not work properly with pagination

### LINKED ISSUE #2579: [Bug] Pagination persists when filtering/searching, showing wrong results
### Directorist version

v8.5.1

### Environment

_No response_

### Description

When I'm browsing listings and navigate to a different page (e.g., page 3), then apply a filter or perform a search, I stay on the same page number instead of returning to page 1. This often results in seeing no results or missing the most relevant results.

**Expected Behavior:**
When I apply a new filter or perform a new search, I should be taken back to page 1 of the results so I can see the most relevant items from the beginning.

**Actual Behavior:**
I stay on the same page number (e.g., page 3), which means:
- I miss the first 2 pages of filtered/searched results
- Sometimes I see "no results" if the filtered results don't have that many pages
- Very confusing user experience

**Additional Context**
Pagination navigation (clicking Next, Previous, or page numbers) works fine - that should keep the page number. The issue is only when performing a new search or applying filters.

### Steps to reproduce

1. Go to the listings page with instant search enabled
2. Navigate to page 3 (or any page other than 1)
3. Apply any filter (category, location, tags, etc.) OR perform a keyword search
4. Observe that you're still on page 3 of the new filtered/searched results

### Isolating the problem

iamibrahimriaz:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
Pagination was persisting across all search actions, causing users to land on incorrect pages when filtering or searching. Added `preservePaged` parameter - `true` only for pagination clicks, `false` (default) for all other searches.

**Before:**
- User on page 3 applies a filter → Stays on page 3 of filtered results
- User on page 5 searches keyword → Shows page 5 of search results
- User on page 4 changes sort → Shows page 4 of sorted results
- Missing first pages of relevant results
- Confusing UX with potentially empty or irrelevant results
- Pagination buttons worked correctly 

**After:**
- User on any page applies filter → Returns to page 1 of filtered results
- User on any page searches keyword → Returns to page 1 of search results
- User on any page changes sort/view → Returns to page 1 of new results
- Always shows results from the beginning
- Clear, expected UX
- Pagination buttons still work correctly 

## Any linked issues
Fixes #2579 

## Checklist


RabbiIslamRony (approved):
Looks Good

PR #2584, opened 2025-11-06
Fix: Hide radius search field when location selector is unavailable

### LINKED ISSUE #2520: Radius search still visible when Location and ZIP fields are not added to the builder
### Directorist version

8.4.8

### Environment

Any

### Description

The radius search option is still visible even when the Location and ZIP fields are not included in the form builder. Since the radius search functionality depends on these fields, it should only appear when at least one of them is added.

### Steps to reproduce

Go to all listing page

**Support Link**
https://taiga-sovware-u10698.vm.elestio.app/project/directorist/issue/1990

### Isolating the problem

iamibrahimriaz:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
Previously, the radius search field would remain visible even when there were no location input elements on the page. This could lead to a confusing user experience where users see a radius search option without any corresponding location field to use it with.

This PR improves the radius search field visibility logic in the search form to handle cases where the location selector elements don't exist on the page.

**Before:**
- Radius search field remains visible even though there's no location selector input field
- Users see a radius search option they cannot use
- Confusing UX with orphaned radius search field

**After:**
- Radius search field is automatically hidden when no location selector elements exist
- Clean UI with only relevant search fields displayed
- Improved user experience

## Changes Made
- Added a check to verify if radius search item selector elements (`$(radius_search_item_selector)`) exist before attempting to loop through them
- When no elements are found, the radius search containers are now hidden by setting `display: none`
- Maintained the existing functionality that hides the radius search when the location field is empty

## Any linked issues
Fixes #2520 

## Checklist


RabbiIslamRony:
Looks Good

RabbiIslamRony (approved):

PR #2628, opened 2025-12-22
fix: Unable to Replace Default Input Label with Custom Label

### LINKED ISSUE #2615: Bug: Unable to Replace Default Input Label with Custom Label
### Directorist version

8.5.6

### Environment

PHP version 7.4.33

### Description

**Description:**
When I remove the text from an input field to add a custom label, the default label keeps reappearing. Even after clearing the field multiple times, the default text restores automatically. To apply a custom label, I have to manually select the field and type again — which creates a poor user experience.

A smoother flow is needed so that when a user clears the label, it stays empty and allows custom text without forcing the default label again.

**Expected Behavior:**
- When the user removes the existing label text, it should remain empty.
- The default label should not return automatically.
- The user should be able to enter a custom label directly.

**Actual Behavior:**
- Removing text restores the default label immediately.
- Requires manually selecting the field again to type a new label.

**Attachment:**
- [Video showing the issue.](https://www.loom.com/share/235d57e90aa24477a0fbfd32cb469be2)

### Reproduction Steps:
1. Go to the Builder.
2. Add a **Number** field in the *Add Listing* form.
3. Save and go to the Add Listing page.
4. Try to edit the label text directly **without selecting the text**.
5. Remove the label text — it immediately restores the default label.

### Isolating the problem

iamibrahimriaz:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [x] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
This PR updates the form builder widget label behavior so that the default label is only applied when the input loses focus (blur) and remains empty, instead of being applied immediately when the user clears the field.

**Before:**
- When a user cleared the label field, the default label was immediately reapplied, making it impossible to temporarily keep the label empty while typing a new one. - https://prnt.sc/Ji9lR2NOQDht 

**After:**
- Users can clear the label and type freely without it being instantly replaced. - https://prnt.sc/oBtywsdb5Lck 
- If the user leaves the label field empty and blurs the input, the default label is applied. - https://prnt.sc/KLdAYg8aMs2G 

## Any linked issues
Fixes #2615 

## Checklist


RabbiIslamRony (approved):
Looks Good

PR #2774, opened 2026-03-31
feat(Listings): add map_card_renderer extension point

nazmulhasan103:
## Summary

Two focused additions to eliminate the need for `Directorist_Listings` instantiation in integrations that only need specific listing output.

### 1. `Helper::get_listing_card_image( int $listing_id ): string`

Consolidates the `thumb-card` template and `loop_get_the_thumbnail()` into a single static call. No class instantiation, no `setup_postdata`, no global post state required.

**Improvements over the current approach:**
- Attachment IDs fetched **once**, reused for both front image and blur-background layer (previously fetched twice per render)
- Accepts `$listing_id` directly — no reliance on `get_the_ID()`
- Private `listing_thumbnail_img()` handles single image, swiper carousel, and default image fallback

### 2. `Directorist_Listings::$map_card_renderer` callable property

Allows integrations to replace the map marker card HTML for both OpenStreet and Google Maps without hooking a global template filter or using shared static state.

**Signature:** `callable( int $listing_id, array $opt ): string`

Defaults to `null` — all existing behaviour fully preserved, zero breaking changes.

```php
$listings->map_card_renderer = function( int $listing_id, array $opt ): string {
    ob_start();
    // render custom markup
    return ob_get_clean();
};
$listings->render_map();
$listings->map_card_renderer = null;
```

## Test plan

- [ ] `Helper::get_listing_card_image()` renders identically to the existing thumb-card template (cover / contain / full modes)
- [ ] Blur background mode renders correctly with a single fetch of attachment IDs
- [ ] Swiper carousel renders correctly for listings with multiple images
- [ ] Default image renders when no images are attached
- [ ] `map_card_renderer = null` (default) — no regression in map rendering
- [ ] OpenStreet map: custom renderer output appears in marker popups
- [ ] Google Maps: custom renderer output appears in info windows

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Armanul46:
Why not use a WordPress filter here instead of a callable property?


$content = apply_filters( 'directorist_map_card_content', 
    Helper::get_template_contents( 'archive/fields/openstreet-map', $opt ), 
    $listings_id, 
    $opt 
);
A filter would be the standard WordPress approach — no need to get a reference to the Listings instance, supports multiple callbacks with priority, and works from anywhere (other plugins, themes, functions.php). The callable property limits it to a single renderer and requires direct object access, which isn't how WP plugins typically expose extension points.

PR #2809, opened 2026-04-12
Fix: CSV import fails to download images from extension-less CDN URLs (e.g. logo.dev)

Armanul46:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description

Problem

When importing listings via CSV with image URLs that do not have a recognisable file extension — such as logo.dev URLs like:

https://img.logo.dev/blackriflecoffee.com?token=pk_xxx
The images were silently skipped and listings showed "No Images" after import.

Root Cause

directorist_rest_upload_image_from_url() extracted the filename from the URL path before the query string, producing blackriflecoffee.com. The original guard only checked for the presence of a dot:

if ( strrpos( $file_array['name'], '.' ) === false ) {
The .com in blackriflecoffee.com satisfied this check, so no MIME detection ran. wp_handle_sideload then received a file with a .com extension, which WordPress rejects as an invalid type, causing the attachment to fail silently.

Fix

Instead of checking for any dot, the fix validates that the detected extension actually maps to an entry in WordPress's allowed MIME types. If it does not — whether because the extension is a TLD (.com, .io, .net) or is absent entirely — the file's real MIME type is read from its contents, the invalid suffix is stripped, and the correct image extension (.jpg, .png, etc.) is appended before sideloading.

Impact

Normal image URLs with valid extensions (.jpg, .png, .webp) — no behavioral change
Extension-less URLs — same behavior as before
Domain-like suffix URLs from CDNs like logo.dev — now correctly downloaded and saved to media library
REST API image uploads (categories, locations, users) — no behavioral change

## Any linked issues
Fixes #

## Checklist


RabbiIslamRony:
Looks good

PR #2836, opened 2026-04-27
Add Nearby sort option for location-based listing results

Armanul46:
Nearby Sort Feature

Nearby is a new sort option for location-based searches. It lets users sort listings by the shortest distance from the searched address or zip code.

How It Works

- Nearby appears in the Sort By dropdown only when:
  - the admin enables it in the relevant sort settings
  - the user searches with valid radius/location coordinates
- It works with both:
  - address-based radius search
  - zip-based radius search
- When selected, listings are ordered by nearest distance first.

Settings

- All Listings page:
  Settings > General > All Listings > Sort Options

- Search Result page:
  Settings > Search > Search Result > Sort Options

Nearby is controlled separately in each area, so:
- enabling it in All Listings affects all-listing sort options
- enabling it in Search Result affects search-result sort options

Behavior Notes

- Radius search filters which listings are eligible.
- Nearby sorts those eligible listings by distance ascending.
- If no valid location reference exists, Nearby does not appear in the dropdown.


RabbiIslamRony:
Looks good

PR #2838, opened 2026-04-28
removed help & support menu

Armanul46:
Removed this https://prnt.sc/cyiDJL5rQUUK
Help & Support menu and its functionality


RabbiIslamRony (changes requested):
Looks Good but
**Things that look “leftover” (not breaking, but now dead/stale)**
- References to the old screen id `at_biz_dir_page_directorist-status` still exist (the screen will never be hit now):
  - `includes/classes/class-updater-notice.php` (screen id list)
  - `includes/asset-loader/helper.php` (`Helper::is_admin_page('support')`)
  - `includes/asset-loader/init.php` (admin enqueue branch for `support`)
- Likely unused assets/strings after removing the page:
  - CSS for `.atbd_help_support` in `assets/css/admin-main.css` (+ RTL)
  - JS tab key `atbds_system-status-tab` in `assets/js/admin-main.js` (if nothing else uses it)

PR #2849, opened 2026-05-10
Fix single listing map popup image fallback

Arafat-plugins:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description

This fixes the single listing map popup image not showing the correct listing image when a listing does not have a preview image.

Previously, `Directorist_Single_Listing::map_data()` only checked the preview image and then fell back directly to the default image. Because of that, listings that had gallery images but no preview image still showed the fallback `grid.jpg` image in the single listing map popup.

This change updates the image fallback order in the single listing map data to:

1. Preview image
2. First gallery image
3. Directory default preview image

It also uses `Helper::default_preview_image_src()` for the default image source and escapes the rendered image output properly.

### Steps to reproduce
1. Create or edit a listing with:
   - no preview image
   - one or more gallery images
2. Open the single listing page.
3. In the `Location` section, open the listing marker popup on the map.
4. Notice the popup shows the fallback `grid.jpg` image instead of the uploaded listing image.

### Expected result
If no preview image is set, the single listing map popup should use the first gallery image before falling back to the default image.

### How to test the fix
1. Go to a single listing page that has:
   - no preview image
   - at least one gallery image
2. Open the map popup from the `Location` section.
3. Confirm the popup now shows the first gallery image.
4. Test a listing with a preview image and confirm the preview image is still shown first.
5. Test a listing with neither preview nor gallery image and confirm the directory default image is shown.

## Any linked issues
Support ticket:
- https://taiga-sovware-u10698.vm.elestio.app/project/directorist/issue/2913

Related screenshots:
- Issue: https://prnt.sc/mPMRSbjnJkQ-
- Fix: https://prnt.sc/yX_Nokqddkw7

## Screenshot
- https://prnt.sc/yX_Nokqddkw7

## Checklist

### DIFF
--- a/includes/model/SingleListing.php
+++ b/includes/model/SingleListing.php
@@ -1353,11 +1353,32 @@ class Directorist_Single_Listing {
         $display_favorite_badge_map = get_directorist_option( 'display_favorite_badge_map', 1 );
 
         $listing_prv_img = directorist_get_listing_preview_image( $id );
-        $default_image = get_directorist_option( 'default_preview_image', DIRECTORIST_ASSETS . 'images/grid.jpg' );
-        $listing_prv_imgurl = ! empty( $listing_prv_img ) ? atbdp_get_image_source( $listing_prv_img, 'small' ) : '';
-        $listing_prv_imgurl = atbdp_image_cropping( $listing_prv_img, 150, 150, true, 100 )['url'];
-        $img_url = ! empty( $listing_prv_imgurl ) ? $listing_prv_imgurl : $default_image;
-        $image = "<figure><img src=" . $img_url . " /></figure>";
+        $listing_img     = directorist_get_listing_gallery_images( $id );
+        $listing_type    = directorist_get_listing_directory( $id );
+        $default_image   = Helper::default_preview_image_src( $listing_type );
+        $listing_title   = get_the_title( $id );
+        $img_url         = '';
+
+        if ( ! empty( $listing_prv_img ) ) {
+            $cropped_image = atbdp_image_cropping( $listing_prv_img, 150, 150, true, 100 );
+            $img_url = ! empty( $cropped_image['url'] ) ? $cropped_image['url'] : atbdp_get_image_source( $listing_prv_img, 'small' );
+        }
+
+        if ( empty( $img_url ) && ! empty( $listing_img[0] ) ) {
+            $gallery_image_id = (int) $listing_img[0];
+            $cropped_image = atbdp_image_cropping( $gallery_image_id, 150, 150, true, 100 );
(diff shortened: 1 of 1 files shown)


RabbiIslamRony (changes requested):
Thanks for the fix. The overall fallback flow is moving in the right direction: the code now avoids unconditionally cropping an empty preview image and adds a gallery/default-image fallback path for the single listing map popup.

I found one fallback edge case that should be addressed before merging.

### Issue: stale gallery attachment IDs can still render invalid image URLs

In `includes/model/SingleListing.php`, the gallery fallback accepts any non-empty `url` returned by `atbdp_image_cropping()`:

```php
$cropped_image = atbdp_image_cropping( $gallery_image_id, 150, 150, true, 100 );
$img_url = ! empty( $cropped_image['url'] ) ? $cropped_image['url'] : atbdp_get_image_source( $gallery_image_id, 'small' );
```

If the gallery image ID is stale or the attachment no longer resolves, `Atbdp_Image_resizer::resize()` can return the attachment ID itself as the `url`. That means the popup can render something like this instead of falling back to the default preview image:

```html
<img src="123">
```

Since this PR is specifically fixing image fallback behavior, we should only use cropped/gallery image values when they resolve to a valid URL. Otherwise, the flow should continue to the default preview image.

### Suggested Fix

One possible approach is to validate the cropped URL and the source URL before assigning `$img_url`:

```php
if ( empty( $img_url ) && ! empty( $listing_img[0] ) ) {
    $gallery_image_id = (int) $listing_img[0];
    $cropped_image    = atbdp_image_cropping( $gallery_image_id, 150, 150, true, 100 );
    $cropped_url      = ! empty( $cropped_image['url'] ) && filter_var( $cropped_image['url'], FILTER_VALIDATE_URL )
        ? $cropped_image['url']
        : '';
    $source_url       = atbdp_get_image_source( $gallery_image_id, 'small' );
    $source_url       = filter_var( $source_url, FILTER_VALIDATE_URL ) ? $source_url : '';

    $img_url = ! empty( $cropped_url ) ? $cropped_url : $source_url;
}
```

Then the existing default fallback can handle unresolved gallery images:

```php
if ( empty( $img_url ) ) {
    $img_url = $default_image;
}
```

The same validation would also be useful for the preview image branch, so both preview and gallery fallbacks behave consistently.

### Verification

I checked the touched file locally:

```bash
php -l includes/model/SingleListing.php
vendor/bin/phpcs --standard=phpcs.xml includes/model/SingleListing.php --report=summary
```

Both passed.

RabbiIslamRony:
@Arafat-plugins 
This issue is not actually in the core. Could you please check again and make sure?

Arafat-plugins:
> @Arafat-plugins This issue is not actually in the core. Could you please check again and make sure?
I am checking

PR #2871, opened 2026-06-03
Fix FormGent enquiry page compatibility

devShamim:
## Summary
- Pin React 18 and compatible WordPress package versions for the FormGent enquiry bundle.
- Add pnpm overrides so transitive Gutenberg packages resolve to compatible React 18-era packages.
- Avoid FormGent protected API usage by preparing response fields locally in Directorist.
- Support both FormGent answer component exports and provide a basic fallback renderer.
- Rebuild the FormGent integration JS/CSS assets.

## Root Cause
The enquiry page was loading a Gutenberg/DataViews dependency graph that could resolve incompatible React/Gutenberg primitives, causing React element objects to be rendered as invalid children. The details endpoint also depended on a protected FormGent controller method, and the modal expected a FormGent answer component export that is not present in the installed FormGent build.

## Validation
- php -l includes/classes/class-formgent.php
- pnpm exec prettier --check package.json
- git diff --check on touched files
- Verified the served formgent-integration.js asset no longer contains React 19, react.transitional.element, @wordpress/primitives@4.47, or webpackMissingModule markers.


RabbiIslamRony (approved):
Looks Good

PR #2922, opened 2026-07-15
fix: prevent dashboard action menu scrolling

syedgalib:
## PR Type

- [x] Bugfix
- [x] Improvement
- [ ] Security fix
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
This fixes the user dashboard listing action dropdown being clipped or positioned upward when only a few listings are displayed.

Root cause:
The responsive listing table owns horizontal overflow, which also makes it the vertical scroll container. With a short result set, an action dropdown could extend outside that container. The existing last-row rule also opened the menu upward even when there were too few preceding rows to contain it.

Changes:
1. Reserved enough vertical space in the user dashboard listing table for action dropdowns when only a few listings are displayed.
2. Kept the last action dropdown opening downward for one-to-three-row result sets so it uses the reserved space.
3. Rebuilt the committed public LTR and RTL CSS assets.

How to test:
1. Go to the frontend user dashboard listing table.
2. Test with one, two, and three listings visible.
3. Open the action dropdown on the last listing row.
4. Confirm the dropdown opens downward and is fully visible.
5. Test with a longer listing set.
6. Confirm the existing last-row dropdown behavior still works correctly.
7. Check both LTR and RTL styles.

## Any linked issues
Fixes #

## Checklist


RabbiIslamRony (changes requested):
https://github.com/syedgalib/directorist/pull/8

RabbiIslamRony (approved):
Looks Good

PR #2926, opened 2026-07-16
fix: skip hidden conditional fields during submission

RabbiIslamRony:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
Fixes a listing submission validation issue where required fields hidden by conditional logic were still validated on the backend.

### Directorist version
Reported from support on Directorist Core 8.8.3.

### Environment
- Local verification target: `https://directorist-core.local/`
- Add Listing page is login-gated on the local site, so the browser-level submit flow could not be completed without credentials.
- Code-level validation was done against the current development-based branch.

### Issue
When a listing submission form has required fields controlled by conditional logic, a hidden required field can block submission.

Example:
1. A selector field has Option 1 and Option 2.
2. Option 1 shows required Text Field 1 and hides required Text Field 2.
3. Option 2 shows required Text Field 2 and hides required Text Field 1.
4. Submitting with Option 1 and Text Field 1 filled still returns a required error for hidden Text Field 2.
5. Submitting with Option 2 and Text Field 2 filled still returns a required error for hidden Text Field 1.

### Root cause
The frontend conditional logic hides fields and disables their inputs, so hidden field values are not submitted by the browser. The backend submission validators still looped through every configured form field and applied required validation without evaluating whether the field should be visible for the submitted values.

### Fix
- Added a server-side conditional logic evaluator for submission fields.
- Skipped validation for fields that evaluate as hidden.
- Skipped saving hidden conditional fields, matching the frontend disabled-input behavior.
- Applied the behavior to both submission paths:
  - `Directorist\AddListingForm\SubmissionController`
  - legacy `ATBDP_Add_Listing`
- Preserved the existing `atbdp_add_listing_form_validation_logic` filter behavior.

### How to reproduce
1. Create a listing submission form with a selector field containing Option 1 and Option 2.
2. Add required Text Field 1 and configure conditional logic to show it only when Option 1 is selected.
3. Add required Text Field 2 and configure conditional logic to show it only when Option 2 is selected.
4. Open the Add Listing form.
5. Select Option 1, fill Text Field 1, and leave Text Field 2 hidden/empty.
6. Submit the listing.
7. Before this fix, submission fails because hidden Text Field 2 is treated as required.
8. After this fix, submission passes that hidden field validation.
9. Repeat with Option 2 and Text Field 2 to confirm hidden Text Field 1 no longer blocks submission.

### How to test the changes
1. Verify Option 1 + filled Text Field 1 passes without requiring hidden Text Field 2.
2. Verify Option 2 + filled Text Field 2 passes without requiring hidden Text Field 1.
3. Verify Option 1 + empty visible Text Field 1 still fails required validation.
4. Verify Option 2 + empty visible Text Field 2 still fails required validation.
5. Verify both the current `SubmissionController::submit()` path and the legacy AJAX add-listing path use the same conditional visibility skip.

### Validation performed
- `php -l includes/fields/class-directorist-conditional-logic.php`
- `php -l includes/fields/init.php`
- `php -l includes/classes/class-submission-controller.php`
- `php -l includes/classes/class-add-listing.php`
- `git diff --check`
- Focused PHP probes for:
  - Option 1 shows Text Field 1 and hides Text Field 2.
  - Option 2 shows Text Field 2 and hides Text Field 1.
  - `action: hide` conditional logic.
  - array values, numeric comparison, and nested taxonomy submitted values.
  - nested submitted arrays do not trigger PHP array-to-string warnings.
- Synthetic evaluator performance probe with an intentionally oversized conditional-logic payload stayed within a small request-time cost.

Note: `composer phpcs` could not be completed in this workspace because `vendor/bin/phpcs` is not installed.

## Any linked issues
Fixes #

## Checklist

### DIFF
--- a/includes/classes/class-add-listing.php
+++ b/includes/classes/class-add-listing.php
@@ -15,6 +15,7 @@ defined( 'ABSPATH' ) || die( 'Direct access is not allowed.' );
 
 use Directorist\Helper;
 use Directorist\Fields\Fields;
+use Directorist\Fields\Conditional_Logic;
 
 if ( ! class_exists( 'ATBDP_Add_Listing' ) ) :
 
@@ -257,7 +258,9 @@ if ( ! class_exists( 'ATBDP_Add_Listing' ) ) :
                         continue;
                     }
 
-                    // Removed: should_ignore_category_custom_field check (assign_to feature removed)
+                    if ( self::should_ignore_conditional_logic_field( $field, $posted_data ) ) {
+                        continue;
+                    }
 
                     switch ( $field->get_internal_key() ) {
                         case 'title':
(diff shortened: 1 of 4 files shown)


RabbiIslamRony:
Replaced by #2927 from RabbiIslamRony:fix/hidden-required-conditional-fields so the PR branch is separated from the main repository remote.

PR #2948, opened 2026-08-02
Preserve IIS temporary paths during uploads and directory import

### LINKED ISSUE #1356: Use of `wp_unslash()` breaks file uploads on IIS hosts
Hello,

When using Directorist on Windows IIS hosted WordPress, file uploads are broken due to the use of `wp_unslash()` on lines 577 and 578 of includes/classes/class-add-listing.php. The result is a broken file path to the temp uploaded file. For example, the file `C:\Program Files\PHPv8.1\upload\phpAC39.tmp` loses all the slashes and ends up being `C:Program FilesPHPv8.1uploadphpAC39.tmp`. So when the file gets passed over to `wp_handle_upload()` it fails the validation because the temp_name is not a valid file path.

https://github.com/sovware/directorist/blob/362c2faafcc490377be2fc78954113849fa83bcd/includes/classes/class-add-listing.php#L577
https://github.com/sovware/directorist/blob/362c2faafcc490377be2fc78954113849fa83bcd/includes/classes/class-add-listing.php#L578

Is it absolutely necessary to unslash the temp_name for the uploaded files? If you're really concerned about validating the values in the `$file` array ('name', 'tmp_name', 'type', etc.) I suggest writing validation that is specific for the array and then validate the value for each key instead of the brute-force approach of running the whole set of data through `wp_unslash()`.

You can set up a simple validation using [`array_filter()`](https://www.php.net/manual/en/function.array-filter.php) to remove invalid files. In the example below, `isFileUploadValid()` is run for every uploaded file. If it returns true, the file is added to the resulting array, otherwise it is not, essentially cleaning out the invalid uploads.

```php
/**
 * Filter callback for use with array_filter(). Receives a file array. Ensures it has all the required keys and 
 * that all the values pass the required validations. Any arrays that fail, will be removed, thus that file will
 * not be uploaded
 */
function isFileUploadValid(array $file): bool
{
    // An array of keys that MUST be in the $file array
    $validKeys = ['name', 'type', 'tmp_name', 'error', 'size',];
    
    // Configure the validations you want to run on each key in $file. The keys in this array 
    // match the names of the keys in $file, the values are the names of callbacks you can
    // set up to validate the key's value. They return true if the value passes, false otherwise
    $validations = [
        'name' => 'fileNameIsValid',
        'type' => 'fileTypeIsValid',
        'tmp_name' => 'fileTmpNameIsValid',
        'error' => 'fileErrorIsValid',
        'size' => 'fileSizeIsValid',
    ];
    
    // Loop through the keys and make sure the $file array has all the ones we want
    foreach ($validKeys as $key) {
        if (!array_key_exists($key, $file)) {
            return false;
        }
    }
    
    // unset $key to avoid any cross-contamination from the previous foreach
    unset($key);
    
    // Loop through the validations and apply them to each value in $file
    foreach($validations as $key => $callback) {
        if (call_user_func($callback, $file[$key]) === false) {
            return false;
        }
    }
    
    // do any other validation you deem necessary
    
    // return true if you get this far
    return true;
}

/*
 * Normalizes the $_FILES array. Necessary when multiple files are uploaded under the same input name, i.e.
 * name="my_input[]". The result is a 2D array where the values for each key are arrays too.
 * [
 *     'my_input' => [
 *         'name' => [
 *             'somefile.jpg',
 *             'somefile2.jpg',
 *         ],
 *         'type' => [
 *             'image/jpg',
 *             'image/jpg',
 *         ],
 *         // etc.
 *     ]
 * ]
 * This function returns a 2D array that looks like this:
 * [
 *     'my_input_0' => [
 *         'name' => 'somefile.jpg',
 *         'type' => 'image/jpg',
 *         // etc.
 *     ],
 *     'my_input_1' => [
 *         'name' => 'somefile2.jpg',
 *         'type' => 'image/jpg',
 *         // etc.
 *     ],
 * ]
 */
function normalizeFiles(array $files): array
{
    $returnArray = [];

    foreach ($files as $file => $fileData) {
        foreach ($fileData as $fileKey => $fileKeyValues) {
            if (is_array($fileKeyValues)) {
                foreach ($fileKeyValues as $index => $value) {
                    $fileName = "{$file}_$index";

                    if (!array_key_exists($fileName, $returnArray)) {
                        $returnArray[$fileName] = [];
                    }

                    $returnArray[$fileName][$fileKey] = $value;
                }
            } elseif (is_scalar($fileKeyValues)) {
                if (!array_key_exists($file, $returnArray)) {
                    $returnArray[$file] = [];
                }
                
                $returnArray[$file][$fileKey] = $fileKeyValues;
            }
        }
    }
    
    return $returnArray;
}

// In the event that you have multiple files uploaded, you'll need to normalize the $_FILES array
$filesToValidate = normalizeFiles($_FILES);

$validUploads = array_filter($filesToValidate, isFileUploadValid);

// continue uploading the file to the DB

```

RabbiIslamRony:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description

1. On Windows/IIS, configure a custom File field in a directory and upload a valid file through the asynchronous listing-form uploader.
2. Confirm the raw temporary path, for example `C:\Program Files\PHPv8.1\upload\phpAC39.tmp`, retains every backslash before `wp_handle_upload()` and the upload succeeds.
3. Go to **Directorist > Directory Builder**, import a valid directory JSON file, and confirm the import succeeds without corrupting the uploaded temporary path.
4. Repeat both flows on a Unix host and confirm custom-file upload, directory import, and the existing listing-image REST upload still succeed.
5. Confirm nonce/capability checks, recursive cleaning, size restrictions, supported MIME validation, and WordPress upload validation remain active.

Verification completed:

- The reported Windows temporary path remained byte-for-byte unchanged; the previous `wp_unslash()` behavior reproduced the corruption.
- Unix temporary paths remained unchanged.
- LocalWP directory JSON multipart import succeeded with HTTP 200 and created the test directory.
- LocalWP asynchronous custom-field multipart upload succeeded with HTTP 201.
- LocalWP listing-image REST upload succeeded with HTTP 200.
- Both changed PHP files passed syntax checks and PHPCS with zero errors; the pre-existing warning count remained unchanged.
- `git diff --check` passed, and all temporary database, plugin, directory, field, and upload-file fixtures were removed.

## Any linked issues
Fixes #1356

## Checklist

### DIFF
--- a/includes/classes/class-ajax-handler.php
+++ b/includes/classes/class-ajax-handler.php
@@ -953,8 +953,7 @@ if ( ! class_exists( 'ATBDP_Ajax_Handler' ) ) :
                     $upload_token = isset( $_POST['upload_token'] ) ? sanitize_text_field( wp_unslash( $_POST['upload_token'] ) ) : '';
                     $token_data   = $upload_token ? get_transient( 'directorist_file_upload_' . $upload_token ) : false;
 
-                    if (
-                        empty( $token_data )
+                    if ( empty( $token_data )
                         || ! is_array( $token_data )
                         || (int) ( $token_data['directory'] ?? 0 ) !== $directory
                         || (string) ( $token_data['field_key'] ?? '' ) !== $field_id
@@ -985,7 +984,7 @@ if ( ! class_exists( 'ATBDP_Ajax_Handler' ) ) :
                 }
 
                 $field_id   = sanitize_text_field( $field_config['field_key'] );
-                $fixed_file = ( ! empty( $_FILES[ $field_id . 'async-upload' ] ) ) ? directorist_clean( wp_unslash( $_FILES[ $field_id . 'async-upload' ] ) ) : '';
+                $fixed_file = ( ! empty( $_FILES[ $field_id . 'async-upload' ] ) ) ? directorist_clean( $_FILES[ $field_id . 'async-upload' ] ) : ''; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash -- Unslashing corrupts Windows upload paths in raw $_FILES data.
 
                 if ( empty( $fixed_file ) ) {
                     throw new \Exception( __( 'No file supplied.', 'directorist' ), 400 );
--- a/includes/modules/multi-directory-setup/class-multi-directory-manager.php
+++ b/includes/modules/multi-directory-setup/class-multi-directory-manager.php
@@ -530,7 +530,7 @@ class Multi_Directory_Manager {
 
         $term_id        = ( ! empty( $_POST[ 'term_id' ] ) ) ? absint( $_POST[ 'term_id' ] ) : 0;
         $directory_name = ( ! empty( $_POST[ 'directory-name' ] ) ) ? sanitize_text_field( wp_unslash( $_POST[ 'directory-name' ] ) ) : '';
-        $json_file      = ( ! empty( $_FILES[ 'directory-import-file' ] ) ) ? directorist_clean( wp_unslash( $_FILES[ 'directory-import-file' ] ) ) : '';
+        $json_file      = ( ! empty( $_FILES[ 'directory-import-file' ] ) ) ? directorist_clean( $_FILES[ 'directory-import-file' ] ) : ''; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash -- Unslashing corrupts Windows upload paths in raw $_FILES data.
 
         // Validation
         $response = [


RabbiIslamRony:
The PHPCS job stopped before code checkout because the current development-branch workflow still uses pull_request_target and actions/checkout refuses fork code in that privileged context. No project code or PHPCS command ran. The infrastructure failure is addressed by #2945; this check should be rerun after that PR merges. Both changed PHP files have zero PHPCS errors locally, with the pre-existing warning count unchanged.

PR #2955, opened 2026-08-05
feat(order): add prepaid status

syedgalib:
## PR Type

- [ ] Bugfix
- [ ] Security fix
- [x] Improvement
- [x] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description

1. Confirm `Directorist\Enums\Order\Status::all()` includes `prepaid`.
2. Run Directorist install/update so the `directorist_orders.status` enum is synced with the new status.
3. Create or update an order to `prepaid` and confirm payment records are not updated to an unsupported payment status.

This PR adds a `prepaid` order status for collected-but-not-yet-renewed payment flows and keeps payment status syncing limited to valid payment statuses.

## Any linked issues
Fixes #

## Checklist

### DIFF
--- a/includes/enums/order/status.php
+++ b/includes/enums/order/status.php
@@ -6,6 +6,7 @@ defined( "ABSPATH" ) || exit;
 
 class Status {
     const PENDING   = 'pending';
+    const PREPAID   = 'prepaid';
     const PAID      = 'paid';
     const FAILED    = 'failed';
     const CANCELLED = 'cancelled';
@@ -16,6 +17,7 @@ class Status {
     public static function all() {
         return [
             self::PENDING,
+            self::PREPAID,
             self::PAID,
             self::FAILED,
             self::CANCELLED,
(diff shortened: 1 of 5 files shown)


RabbiIslamRony (changes requested):
The prepaid enum, REST mappings, and order/payment synchronization behavior work as intended in the targeted tests. However, the database migration introduces a blocking schema regression.

sync_order_status_enum() runs ALTER TABLE ... MODIFY status ENUM(...) DEFAULT ..., but the replacement column definition omits NOT NULL. MySQL MODIFY replaces the complete column definition, so this changes directorist_orders.status from NOT NULL to nullable. An isolated local migration test confirmed Null changed from NO to YES and a NULL status value was accepted.

Please preserve the complete status column definition, including NOT NULL and the pending default, and verify both fresh-install and existing-table upgrade paths. The migration should also confirm that existing order data remains unchanged.

Local functional validation on commit 6630660a39 confirmed that prepaid leaves the payment pending and a later paid transition correctly synchronizes the payment.

RabbiIslamRony (approved):
Rechecked the updated migration on commit 2cb30a0a5f. The order status column now preserves NOT NULL with the pending default. Existing-table upgrade and fresh-install tests passed, existing rows remained unchanged, prepaid values were accepted, and NULL values were rejected. The prepaid-to-paid order and payment synchronization flow also passed. Approved.

PR #3025, opened 2026-09-16
Require approval for expired listing renewals and reset expiry on publish

the-sohan:
## PR Type

- [x] Bugfix
- [ ] Security fix
- [ ] Improvement
- [ ] New Feature
- [ ] Build related changes
- [ ] Text changes

## Description
The dashboard Renew action for an expired listing could lead directly to a checkout URL without a valid `checkout_type`, displaying “Invalid checkout type.” The core renewal handler also published listings immediately, leaving no opportunity for admin approval.

Renew now uses the authenticated renewal route and submits the listing as Pending Review. A renewal marker identifies listings awaiting approval. When a pending renewal is published, Directorist calculates `_expiry_date` from the approval time using that listing directory's Default listing expiration days, clears the marker, and fires the renewal completion hook. The owner sees a submission-for-approval notice. The plain-permalink renewal query parameter is also corrected.

Files changed:
- `includes/classes/class-add-listing.php` — submit renewals for review and reset expiration when approved.
- `includes/classes/class-permalink.php` — correct the listing ID query key for plain permalink renewal URLs.
- `includes/model/ListingDashboard.php` — route dashboard Renew through the renewal handler and pass the pending notice state.
- `templates/dashboard/notice.php` — show the owner that renewal awaits approval.

How to reproduce the issue or test the changes:

1. As a listing owner, open an expired listing in Dashboard → My Listings → More → Renew. Before the change, the flow may show “Invalid checkout type” or publish immediately.
2. With this change, confirm the listing becomes Pending Review and the owner sees “Renewal submitted for approval.” Its old expiration date must not be extended yet.
3. As an administrator, publish the pending listing. Confirm the new expiration is the approval time plus the listing directory's Default listing expiration days and that a subsequent edit does not renew it again.
4. Check that an ineligible listing cannot be renewed through the renewal URL.

Local verification:

- On the local site, the owner's Renew action moved an expired listing to Pending Review while leaving its old expiration unchanged.
- Publishing that pending listing in wp-admin set expiration seven days from approval, matching the directory default.
- Rechecked the approval hook with a pending-to-published transition after switching to `wp_insert_post`; it calculated seven days and cleared the renewal marker. The test listing was restored to its prior expired state.
- `php -l` passed for all four changed PHP files; `git diff --check origin/development...HEAD` passed.

Notes:

- Paid featured listing and Pricing Plans extension renewal flows were not exercised on the local site.
- The repository PHPCS executable is unavailable because Composer dependencies are not installed in this worktree.

## Any linked issues
- Help Scout: https://secure.helpscout.net/conversation/3427661919/6339?viewId=8936453
- Customer video: https://drive.google.com/file/d/1S9MUDTwh2DQyDM5JNmeB31hR2r16xc6L/view?usp=sharing

## Checklist


syedgalib (changes requested):
Changes requested. The reported checkout error is fixed, and the direct renewal, approval, ineligible-listing guard, and plain-permalink paths work. I found two regressions in the renewal lifecycle:

1. **[P1] Preserve renewal state when a submission is rejected** (`includes/classes/class-add-listing.php:1059`). This branch deletes `_directorist_renewal_pending` on `rejected`. An owner can edit a rejected listing, and `directorist_get_listing_edit_status()` sends that edit back to `pending`. When an administrator then publishes it, `complete_renewal_approval()` exits because the marker is gone, leaving the original expired `_expiry_date`. The listing is published with an expiration in the past and can be expired again by the expiration job. Keep the marker through resubmission, or otherwise recognize that it is still a renewal when approved.

2. **[P2] Refresh the listing date on renewal approval** (`includes/classes/class-add-listing.php:1064`). The previous renewal path set both `post_date` and `post_date_gmt` to the renewal time. This path only changes status and expiration, so an older renewed listing keeps its original date. The dashboard and listing results use date sorting, which leaves renewed listings at their old position instead of the position they had before this change. Update the publication date when approval completes if renewals should retain their existing sort behavior.

the-sohan:
Addressed both requested changes in 694f24726:

- Preserved `_directorist_renewal_pending` when a renewal is rejected, so owner resubmission remains identifiable and approval resets the expiration correctly. The marker is still removed when the listing is trashed.
- Refreshed `post_date` and `post_date_gmt` at approval time and used that same timestamp as the start of the new directory-default expiration period.

Verified locally with `pending → rejected → pending → publish`: the marker remained through rejection/resubmission, was cleared on approval, both publication dates were refreshed, and a 7-day directory default produced an expiry exactly 7 days after approval. The test listing was restored afterward.

syedgalib (approved):
Approved. The new commit resolves both issues from my previous review. Renewal still moves an expired listing to Pending Review without extending its expiration early; the renewal marker survives rejection and a return to pending; approval refreshes the publication date and expiration from the directory default, then clears the marker. A later edit does not renew it again, and an approved listing cannot be renewed through the old URL. The plain-permalink and checkout routing changes remain sound. I found no remaining blocking issue in the reviewed paths.

